PR #2, opened 2020-01-10
Use content type templates

rlskoeser:
Template restructuring and other small changes. Files have been moved around and renamed slightly, but everything should render as it did before when building the site.

- Restructure templates to use type-specific layouts to avoid having to specify both type and layout in the front matter, as much as possible.
- Configure a permalink for all content under `issues/` that uses full section path and slug or title instead of filename, which I think gives us more flexibility for our urls
- Revise issue template to use `.Sections` (this is simpler and probably more efficient; fixes a small issue where it was listing itself with almost no data).


content/issues/1/_index.md
```diff
@@ -1,10 +1,11 @@
 ---
-layout: issue
 type: issue
+layout: single
 title: Issue One
-number: One
+number: 1
```
rlskoeser: I meant to look where this was used and forgot. I think we should store as a numeral so we can sort if necessary (although I suppose we could always sort on publish date). I expect there should be a way to generate the text version from the numeral in Go templates; willing to look into that if it's helpful/important.

gwijthoff:
So much more elegant. Thanks @rlskoeser

PR #3, opened 2020-01-10
Configure text output for articles

rlskoeser:
- configure text as an output format
- add text output templates (a generic one and an article specific one)
- link to all alternate formats in html page head
- link to text alternate format at the bottom of the page on articles

This includes changes in #2 and should be reviewed after that has been merged.


gwijthoff:
Plain text output makes this thing feel so much closer to real

PR #46, opened 2020-03-19
semanticize html & reset styles

thatbudakguy:
this PR:
- restructures the HTML for most pages to use a more semantic structure
- strips out all tachyons styles
- reworks the style pipeline to allow partials
- adds a few very basic styles for layout purposes
- changes the available partial templates
- removes some unused parts of the default theme


content/about.md
```diff
@@ -3,22 +3,28 @@ layout: page
...
-> (computer science) A string of letters that identifies the beginning of a valid sequence in a specified language.
-
-> (literature) The first appearance of a genre in print.
-
-<p class="tc">❀</p>
-
-*Startwords* is a research periodical irregularly published by the Center for Digital Humanities at Princeton. More formal than a blog yet more speculative and process-based than a peer-reviewed journal, *Startwords* is a forum for experimental humanities scholarship. Embedded code, data physicalizations, and emerging forms of project documentation are explored through writing that is creative, conversational, exploratory, essayistic, data-driven, and academically rigorous.
-
-Each issue features two articles that are unified by a common theme, feature, or concern. Issues will be released one to three times a year, when we have something to share or say. Articles can be contributed by CDH staff, faculty, graduate students, or external collaborators.
+<dl>
```
rlskoeser: There is markdown syntax for definition lists, and it seems to be supported by whatever processor we're using.
Seems like it would be more readable to use markdown, and that we should prefer markdown to html where possible.
https://www.markdownguide.org/extended-syntax/#definition-lists

config.toml
```diff
@@ -1,10 +1,8 @@
 baseURL = "http://startwords.cdh.princeton.edu/"
-languageCode = "en-us"
```
rlskoeser: Any reason this is removed?

content/issues/1/_index.md
```diff
@@ -3,7 +3,7 @@ type: issue
 layout: single
 title: Issue One
 number: 1
-issue_theme: Transformations
+theme: Transformations
```
rlskoeser: I *think* that @gwijthoff and I used `issue_theme` instead of `theme` here to avoid confusion with Hugo theme settings, but I don't know if that actually matters in the context where we use this.

themes/startwords/assets/scss/main.scss
```diff
@@ -0,0 +1,4 @@
+// Startwords stylesheet
+@import 'global';
+@import 'a11y';
+@import 'print';
```
rlskoeser: Do we want this included in the main scss?  I had imagined it as a separate include so we can specify media, but I guess we can scope to print media within the print.scss file?

themes/startwords/theme.toml
```diff
@@ -1,12 +1,12 @@
-name = "Blank"
+name = "startwords"
 license = "MIT"
-licenselink = "https://github.com/vimux/blank/blob/master/LICENSE"
-description = "Starter Hugo theme for developers."
+licenselink = "https://github.com/Princeton-CDH/startwords/blob/master/themes/startwords/LICENSE"
+description = "Theme for the Startwords research periodical."
 homepage = "https://github.com/vimux/blank/"
-tags = ["blog", "plain", "blank", "starter", "development"]
+tags = ["journal", "research", "periodical"]
```
rlskoeser: Thanks for updating these!

rlskoeser: It looks like we should update the license, it still lists vimux as author.  

The theme license is MIT, can we / do we want to change that?

I don't think we actually have a license on the codebase yet, but we should probably add one.

themes/startwords/theme.toml
```diff
@@ -1,12 +1,12 @@
-name = "Blank"
+name = "startwords"
 license = "MIT"
-licenselink = "https://github.com/vimux/blank/blob/master/LICENSE"
-description = "Starter Hugo theme for developers."
+licenselink = "https://github.com/Princeton-CDH/startwords/blob/master/themes/startwords/LICENSE"
+description = "Theme for the Startwords research periodical."
 homepage = "https://github.com/vimux/blank/"
```
rlskoeser: Is this the theme homepage? Should it point to the theme within the startwords repo?

themes/startwords/theme.toml
```diff
@@ -1,12 +1,12 @@
...
-tags = ["blog", "plain", "blank", "starter", "development"]
+tags = ["journal", "research", "periodical"]
 features = ["blog"]
 min_version = "0.20"
 
 [author]
-  name = "Vimux"
-  homepage = "https://github.com/vimux"
+  name = "Center for Digital Humanities @ Princeton"
+  homepage = "https://github.com/Princeton-CDH"
```
rlskoeser: I wonder if author homepage should be cdh.princeton.edu instead?

themes/startwords/layouts/article/single.html
```diff
@@ -1,46 +1,16 @@
...
-</div>
-
-<br/>
-<br/>
-
-<div class="pl7 f4 mw7">
-	<p class="tc">❀</p>
-	{{ .Content }}
-</div>
+{{ define "pagetype" }}article{{ end }}
```
rlskoeser: I wonder if you could infer page type from content type where we have one?

config.toml
thatbudakguy: I couldn't imagine a scenario where the site's language code would be anything other than `en-us`; even if we were to publish a full article in another language we would need to scope a `lang=` attribute to the `article` in question. it didn't seem like something worth storing in a config.

content/issues/1/_index.md
thatbudakguy: as far as I could tell there were no repercussions. I think `theme` is stored in `.Site.Theme` whereas this would be `.Params.theme` or something, but if it causes conflict elsewhere we can switch it back.

themes/startwords/assets/scss/main.scss
thatbudakguy: we can scope it, but you're right - this reminded me that we probably do want it in a separate file so that we only include it when the user prints the page, to save space. there's a way to do this in the `link` tag, iirc.

rlskoeser: Yes, that's what I had in mind.

themes/startwords/theme.toml
thatbudakguy: oh, good catch! not sure what it should really do since I guess our theme isn't published on GitHub on its own. I guess we could also just remove it?

themes/startwords/theme.toml
thatbudakguy: it definitely should, good idea!

themes/startwords/theme.toml
rlskoeser: yeah, I like that idea even better

themes/startwords/layouts/article/single.html
thatbudakguy: that would be amazing, especially if it could differentiate between single and list views! I also thought about using the template name or path somehow (`issue/list`) but that's beyond my very limited Hugo knowledge so far...

themes/startwords/theme.toml
thatbudakguy: let's remove the theme license (since we don't intend to distribute it independently). maybe we can use Apache2 for the main codebase, since we use that on most other projects.

PR #52, opened 2020-03-23
Add deepzoom shortcode with openseadragon

rlskoeser:
- defines new deepzoom shortcode (could add more customization options but I'm not sure what we need yet)
- preliminary style for deepzoom display div
- include openseadragon js on pages that use the deepzoom shortcode
- use deepzoom shortcode in data beyond vision article

Should we add a readme in the shortcode directory where we can list and document our local shortcodes and how they can be used/invoked?


themes/startwords/layouts/_default/baseof.html
```diff
@@ -22,6 +22,11 @@
     {{ range .AlternativeOutputFormats -}}
     <link rel="{{ .Rel }}" type="{{ .MediaType.Type }}" href="{{ .Permalink | safeURL }}">
     {{ end -}}
+
+    {{ if .HasShortcode "deepzoom" }}{{/* include openseadragon when deepzoom shortcode is used */}}
```
thatbudakguy: this is great; love the ability to conditionally include it.

thatbudakguy:
re: documenting the shortcodes, it seems like this could be part of our editorial workflow/documentation for contributors, wherever that will eventually live.

rlskoeser:
I think if template/shortcode documentation is not in the github repository we're less likely to maintain it.

Issue #15, opened 2020-03-16
As a contributor, I want to embed images in pages.


thatbudakguy:
# dev notes
we need to support:
- [x] embedded (accessible) images
- [x] icons (as hyperlinks)
- [x] custom background images
# testing notes
there are three parts to this issue; two of which are already accomplished:
1. we can use hugo's built-in [`figure` shortcode](https://gohugo.io/content-management/shortcodes/#figure) to embed images with accessible captions.
2. we can create image/icon links in markdown that point to headings using this format:
```md

## My Heading {#link-name}

<!-- this will make an image link that points to "my heading" -->
[![description of image](path/to/image/file.jpg)](#link-name)
```
for the third part (background images), this was implemented by allowing custom css to be applied to any article or issue page. this way, we can use a css rule to import any image and set it as the background for an article. you can see an example on the data beyond vision essay, which loads the `weaving.jpg` image and adds it to the article background. here's how to test this feature:
## steps
1. pull an updated copy of the master branch and switch to it:
```sh
$ git pull origin master
$ git checkout master
```
2. you'll notice that article markdown files are now all named `index.md` and contained within their own folder. images and other content specific to that article can go together in the folder to form a [page bundle](https://gohugo.io/content-management/page-bundles/#examples-of-leaf-bundle-organization).
3. inside any page bundle, create a new `.css` file with any name you like (I used `style.css`). it should be at the same hierarchy level as the article's `index.md`. you can see an example one under the `data-beyond-vision/` folder.
4. add a rule to the css file with something that will be immediately visible, to test it out - i used:
```css
article { background: red; }
```
5. restart your hugo server and view the article. you should see the effects of the css (a glaring red background!)
6. try repeating steps 3-5, but at the _issue_ level, creating your css file next to the `_index.md` for that issue. you can see the effect on the issue page.

if everything works as you expect, feel free to close the issue.

thatbudakguy:
NOTE break this up into smaller pieces so it's testable

gwijthoff:
Looks great, thanks for your clear testing notes here @thatbudakguy

PR #55, opened 2020-03-24
image embedding modes

thatbudakguy:
refs #15 

this reorganizes article content so that each article is in its own folder in order to take advantage of page bundles, and allows for custom css that can be applied to articles or issues by adding `.css` files to their folder.

it also adds image assets for the data beyond vision essay with examples showing how to make image links to section headings and a background image via css.


themes/startwords/layouts/article/single.html
```diff
@@ -1,3 +1,9 @@
+{{ define "css" }}
+    {{ range .Resources.Match "*.css" }}
+    <link rel="stylesheet" href="{{ .RelPermalink }}">
+    {{ end }}
+{{ end }}
+
```
rlskoeser: This is a great solution! We should make sure to document how to add page-specific css somewhere (readme?).

I wonder if this could be added to the base template so that it would apply to any page and not have to be in both article and issue layouts?

thatbudakguy: I wondered about that too, but wasn't sure what the actual implications would be...for instance, if I dropped a `style.css` file into the entire `content/` root...what would happen? It seemed like there might be a potential for confusing behavior to occur, and I also didn't make any provisioning for ensuring that multiple files cascaded in the correct order.

rlskoeser: I thought that the resources were specific to page based on our experience with the images earlier, but I think you've worked with this more and I trust your inclination.

PR #61, opened 2020-04-14
styles/homepage

thatbudakguy:
this PR implements the mobile and tablet styles for the homepage.


content/_index.md
```diff
@@ -3,4 +3,7 @@ layout: page
 title: Startwords
 ---
 
-Read more about the project on the [about page](/about).
+startword, <abbr title="noun">n.</abbr>
+: (computer science) A string of letters that identifies the beginning of
+    a valid sequence in a specified language.
+: (literature) The first appearance of a genre in print.
```
rlskoeser, line 9: Is there a way to use the about page summary here instead of duplicating the content? (probably not high priority, doesn't need to be addressed for this PR)

themes/startwords/assets/scss/_utilities.scss
```diff
@@ -0,0 +1,32 @@
+/* utility mixins & functions */
+
+// inverted sections use a purple background & white text
+@mixin inverted {
```
rlskoeser, line 4: nice

themes/startwords/assets/scss/_utilities.scss
```diff
@@ -0,0 +1,32 @@
...
+    @include inverted;
+}
+
+// convert a pixel measurement into rems
+// '100px' @ 1rem = 18px -> ''
+@function rem($px) {
+    $rem: $px / $base-font-size;
+    @return #{$rem}rem;
+}
+
```
rlskoeser, line 24: cool! look forward to using this

themes/startwords/layouts/article/summary.html
```diff
@@ -1,5 +1,7 @@
 <article class="article-summary">
-    {{ .Summary }}
-    <h2><a href="{{ .Permalink }}">{{ .Title }}</a></h2>
+    <a href="{{ .Permalink }}">
+        <p>{{ .Summary | plainify }}</p>
```
rlskoeser, line 4: Why do we need `plainify` here ?

themes/startwords/layouts/index.html
```diff
@@ -1,15 +1,29 @@
...
+    <div class="container">
+        {{ range first 2 (where .Site.AllPages "Type" "article") }}
+        {{ .Render "summary" }}
+        {{ end }}
+    </div>
+    {{ with .Content }}
+    <div class="content">
+        {{ . }}
+        <a class="more" href="/about">
+            Read more about <em>Startwords</em> here.
```
rlskoeser, line 25: Can we drop the "here" from this text?

thatbudakguy:
yeah, it's because there's now a grid applied to `<main>` but it's children aren't styled on any page except the homepage yet.

themes/startwords/layouts/article/summary.html
thatbudakguy: without it, things like footnotes will be rendered as `<a>` here, and we never want to render anything other than plaintext (this was actually happening with the body as data essay)

themes/startwords/layouts/index.html
thatbudakguy: definitely! makes sense to me

content/_index.md
thatbudakguy: I don't think I know how to select the about page to get its summary here, but that makes sense to me!

Issue #60, opened 2020-04-14
As a reader, I want to navigate around the site using a custom navigation in the header.


thatbudakguy:
# testing notes

- the hiding/"pitbar" behavior for this isn't implemented yet; that will be tracked on #66.
- it should look the same on all pages.
- it should always be visible.

# dev notes
this is the menu that appears vertically along the top left portion of the page on all pages. it always contains links to the homepage and the issues list page.

gwijthoff:
Is the home & issue button/links showing up like this in anyone else's browser? It's the case for me on my iPhone and my tablet in both Firefox and Safari. Looks as designed on my laptop though. 

![IMG_5469.jpg](https://images.zenhubusercontent.com/5e6fa2284847cbacb75fdd35/d403f92f-05b6-4a8b-970a-e2eafcaefbe0)

thatbudakguy:
@gwijthoff I tested on an iPad just now and it looked OK in both safari and firefox. this might be a cache problem - this issue was happening earlier and I deployed a fix, so maybe you have an old version cached. try clearing your data?

gissoo:
@thatbudakguy This is still happening to me only on Safari now that I check for the first time.

rlskoeser:
@thatbudakguy I was able to replicate with Safari on an iPad. IDK why it's needed, maybe older devices have older versions of the browsers? 

@gwijthoff @gissoo I added a webkit specific variant of the clip-path that's used for the nav shapes. It's on the site, so you can review and confirm it's working on your devices now.

gwijthoff:
@rlskoeser that fixed it for me! Home & issue button/links now look as they should on my phone and tablet.

thatbudakguy:
@rlskoeser thanks for doing that!

out of curiousity, could @gissoo or @gwijthoff check your version of iOS safari? compat tables say that the vendor prefix hasn't been needed since version 9, and the current version is 13, so would be good to know if some devices are indeed that far behind.

gwijthoff:
@thatbudakguy I was using Firefox 24.1 (17574) on an iPhone 6s running iOS 13.3.1

thatbudakguy:
huh... the compatibility tables don't even list anything for Firefox on iOS. the Firefox on Android current version is 68, so I guess Firefox on iOS is on a totally different versioning schedule. thanks for checking!

gissoo:
[...]

PR #62, opened 2020-04-14
implement header navigation (#60)

thatbudakguy:
this implements the site's main navigation (the home & issue button/links).


themes/startwords/layouts/partials/header.html
```diff
@@ -1,8 +1,16 @@
 <header>
     <nav aria-label="main">
         <ul>
-            <li><a class="issues" href="/issues">Issues</a></li>
-            <li><a class="home" href="/">Home</a></li>
+            <li class="issues">
```
rlskoeser, line 4: Would it make sense to configure the menu in the site config ? You can see the footer for an example

thatbudakguy: we could do this, but there seem to be at least two reasons not to:
- theoretically this menu won't ever change (changing it would at least require new design) and such a change would be pretty major in terms of content architecture
- one of the two items needs to be "special" because it's an image with some accessible text rather than just text anyway

rlskoeser: Right - it wouldn't make sense for the logo to change, and it doesn't get us a lot to make the 'issues' part of the menu configurable.

thatbudakguy:
looks like chrome doesn't support the `path()` value for `clip-path()`s, but luckily I was able to find this incredibly simple [path to polygon converter](https://betravis.github.io/shape-tools/path-to-polygon/)! looks like it's working in chrome now.

rlskoeser:
Looks great!

PR #63, opened 2020-04-15
Footer icons and styles

rlskoeser:
Revise footer content and adds styles to match the design.

I'm using the black SVG logos as masks so the color can be controlled in CSS (haven't removed the white SVG logos yet, but I think we can after review if you're ok with this approach). I opted to make the logos off-white instead of white because the white seemed kind of glaring in comparison to the off-white link text, but I want to make sure Gissoo is ok with that when she reviews.

I'm not sure I have the positions / spacing exactly right, but have been fiddling with it enough I wanted another set of eyes on it.

The homepage doesn't currently match the design exactly - I looked at it some, but I wasn't sure the best way to have the off-white page section extend under the footer.


thatbudakguy:
hey, documenting my thoughts re: how to get the right coloring on the homepage. i talked with @gissoo some about the logic behind `.inverted` to make sure it meshes with her concept of pages - here's how that breaks down.
- homepage: **inverted** (background color is purple, footer is dark purple)
- issue list page: **normal** (background color is white, footer is gray)
- issue detail page: **inverted** (special white section at top, but overall background is purple, footer is dark purple)
- article page: **normal** (background color is white, footer is gray)

the homepage is a special case because it has a white section at the _bottom_ by the footer, which causes problems because we want the footer to appear to be on a white background. my proposed solution for this (on the homepage only) is to:
- make the white section at the bottom of the page (`.div.content`) have a large `padding-bottom` value (equal to the height of the footer plus some extra)
- keep the `clip-path` on the footer so that it has no background
- "pull" the footer upwards over the white section using a negative `margin-top` value.

Issue #14, opened 2020-03-16
As a reader, I want to read notes in context as I'm reading an article so that I can refer to additional information without losing my place.


thatbudakguy:
## testing notes
- notes should be full width up to 414px
- when notes are full width, they should have a slide up/down transition
- when notes are not full width, they should be position so the arrow is under the reference and they should fade in/out
- notes should disappear automatically when you scroll
- if you click on a reference near the bottom of the page and the note is not fully visible, the page should scroll up so you can see the note; the note position should adjust to match the scroll
- on desktop, if the reference is close to the left side of the page the triangle shape should be flipped (to ensure there is space for arrow & reference to align) [might have to adjust browser window size to get a reference so it just wraps)
- clicking the x icon on an open note should close the icon and not jump you to the reference
- endnotes should be visible when you scroll all the way down; backlinks should take you to the reference in the text
- if you load the page with a note id in the url, the note should display 

--- 

# dev notes
[figma link](https://www.figma.com/file/MEt3gfKX4N3WTLPbLGBbGC/Startwords?node-id=153%3A2766)
this corresponds to the ability to add footnotes, contextual notes, etc - however we implement that. it could involve a shortcode, as suggested in #4.

## revisions:
- [x] full width up to 414px
- [x] change [reference link underline style](https://www.figma.com/file/MEt3gfKX4N3WTLPbLGBbGC/Startwords?node-id=686%3A0) – 5px wide, with a 60% transparency with gradient
- [x] always show reference underline, not just on hover
- [x] fix number alignment under point of arrow
- [x] fix outline display issue on safari
- [x] fix link color
- [x] fix note position on mobile
- [x] override close button so document doesn't scroll to reference anchor
- [x] slide up/down transition on mobile
- [x] fade in/out transition on tablet/desktop
- [x] swap arrow icon for x icon
- [x] set class and styles for notes when displayed as end notes

rlskoeser:
Currently implemented with standard markdown footnotes. There are two examples in the current Data Beyond Vision article draft.

rlskoeser:
I noticed that the text version of the articles includes some junk at the end of the footnotes — I think this is the return to text link generated by the markdown tool. Maybe we can just configure markdown not to generate one, since I don't think we need it based on Gissoo's design.  I spent a little time trying to remove it with a "replace" call in the template but it didn't work and I decided it made sense to hold off until we implement this.

FWIW, I think this issue title is too vague

[...]

PR #65, opened 2020-04-15
Initial contextual note implementation (CSS + JS positions, no transitions)

rlskoeser:
Adds styles on top of footnotes that are generated by standard markdown footnotes to display them as contextual notes when they are the `:target` element (i.e. reference has been clicked).

The return link is positioned and styled to be used as the down arrow.

Additional refinement will require javascript, but this version is at least semi-usable and I thought it would be useful to get feedback on.

If we want to do a slide up on mobile we'll have to use some other styles, since you can't apply transition to position. (Maybe height? 0 by default, with js to set a class on the notes container when it's visible to set height back to auto.)


thatbudakguy:
this works and looks pretty good on mobile.

on tablet, long footnotes are cut off and it looks awkward because the container is mobile-sized and doesn't fit the screen. could we stretch it to fit the screen width, so the notes would be much shorter but still take up the lower part of the screen?

i wonder about the interactivity of the down arrow. personally, I would expect that to just close the footnote (in the same way you would close a popup) rather than also taking you back to the reference. what if we made the number inside the footnote link back to the reference, and had the arrow just close the popup?

i think we might need to explore making the popups non-fixed on desktop, since there's no way we can conceivably stretch one large enough to fit the screen, and it looks strange to have them hovering there covering other content if they don't fill the screen.

thatbudakguy:
just noticed that the numbering of the endnotes is off also...not sure what to do about that.

rlskoeser:
@thatbudakguy I don't know if you saw, I wrote up some notes from my conversation with @gissoo earlier this week in a comment on this issue https://github.com/Princeton-CDH/startwords/issues/14#issuecomment-614199877 

I had originally imagined the notes would be full-width on tablet too (extrapolating based on the footer, I think), but that isn't what @gissoo had planned.

thatbudakguy:
I hadn't seen that, thank you! this explains a lot. Since the only situation we would need to handle a lack of js is on desktop, I think this is functional enough that a non-js user could at least see the right content, so I'm good with it.

rlskoeser:
@thatbudakguy I was going to ask where the numbers were off and then thought I'd better check Firefox. Chrome isn't doing that. I wonder if it's something I did with the counter... But I can investigate that. (If need be, I know FF supports `::marker` !)

thatbudakguy:
we could just also force restart numbering using the `counter-reset` attribute like we do in s&co, I think

rlskoeser:
@thatbudakguy I think I've fixed the endnote counter issue, please review.

I forgot to ask: on mobile the contextual note isn't full width, but I'm having trouble figuring out why. Is there something in the layout further up that's causing the thin white stripe on the right?

Do you want to merge the css-only notes (when it's satisfactory) or wait until we have the js implemented also?

rlskoeser:
@thatbudakguy updated with preliminary javascript to position contextual notes relative to references on tablet/desktop, and hide them on scroll. Would appreciate your input!

themes/startwords/assets/js/notes.js
```diff
@@ -0,0 +1,58 @@
+/* javascript for notes behavior */
+
+window.onload = function() {
+	// position contextual note if the page is loaded with one selected
+	// FIXME: not displaying on load - is this due to scrolling ?
+	positionContextualNote();
```
thatbudakguy, line 6: you could try using the `document.addEventListener('DOMContentLoaded')` hook instead; that's guaranteed to fire before css takes effect and might have different results. we use that when setting up search pages in s&co to avoid tearing/visual movement of things.

themes/startwords/assets/js/notes.js
```diff
@@ -0,0 +1,58 @@
...
+};
+
+window.addEventListener('hashchange', function() {
+	positionContextualNote();
```
thatbudakguy, line 10: this is an interesting hook i didn't know about. we might need to make sure we don't load this js on pages other than the article page, because some pages (like the issue detail page) use anchor links to move the user around (like when you click the chevron) but have no relationship to contextual notes. although I guess the function could fire and do nothing...

themes/startwords/layouts/_default/baseof.html
```diff
@@ -27,6 +27,11 @@
     <link rel="stylesheet" href="{{ $print.RelPermalink }}" media="print">
     {{ block "css" . }}{{ end }}
 
+    {{ $notes := resources.Get "js/notes.js" }}
+    {{/*     {{ $global := resources.Get "js/global.js" }} */}}
```
thatbudakguy, line 31: is this line commented? not sure what it is/means

rlskoeser: Sorry - commented out based on the example from the hugo docs; I left it in because I thought we might have a main.js at some point. Should I remove it until we do?

themes/startwords/assets/js/notes.js
rlskoeser: I had a question about whether contextual notes should work on any pages or only on articles - in theory we could have markdown notes on other content pages, but IDK if we'd want them to behave like this or not.

The function looks for fn style hashes and should ignore everything else.

thatbudakguy: good point - should be safe to leave it in, then.

themes/startwords/layouts/_default/baseof.html
thatbudakguy: ah, got it - i guess there's no harm in leaving it since we should soon have a `main.js` with the logic for the main nav.

rlskoeser:
@thatbudakguy updated with logic to scroll the page if the full note is not visible, and temporarily disable hiding notes on scroll in that case and also on page load. I'm able to link to a page with a note and have it visible now (tested both FF and Chrome).

thatbudakguy:
linking to specific footnotes is working for me, too! nice!

Issue #17, opened 2020-03-16
As a reader, I want to browse a list of all published issues.


thatbudakguy:
# testing notes
[website link](https://startwords.cdh.princeton.edu/issues/)
[figma link](https://www.figma.com/file/MEt3gfKX4N3WTLPbLGBbGC/Startwords?node-id=153%3A1707)
this is the issue list page.

one note for design testing - as @rlskoeser noted, the design shows the red arrow appearing when each issue is hovered (but otherwise invisible). I have implemented it more in line with the red arrow elsewhere (e.g. on the homepage) - the arrow is always visible, but shifts subtly on hover. I can change it if that doesn't make sense here.

rlskoeser:
Layout for a list of  issues is in startwords theme under layouts/issue/list.html and includes logic to loop over and link to all issues, and display the issue theme and content.

gissoo:
@rlskoeser 
- I think the space between "Issue number" and "issue date" should be decreased to be 7px and the space between issue date and the theme should be 14px – sorry that Figma is not making this easier  

- Would it be possible that for all mobile the width of the content on this page would remain as is and not get wider? – I seriously think it will look so much better – but not sure if this is disastrous systemically.

thatbudakguy:
I've adjusted the spacing between "issue number", "issue date", and the theme.

For the resizing - it's possible, but I'm not sure how to make it interact with our grid. I have defined the content container for this page (and others) to span from grid column 2 to column 9 on mobile, and from column 4 to column 7 on tablet, since that's what's depicted in the design. If the device is a different size from our breakpoint, the grid will be scaled to fit it, so the actual size of the content may change, but it will always take up the same amount/percentage of screen space. Both pictures below are inside our mobile breakpoint; one is ~375px wide (iPhone) and the other is ~700px wide. I don't think there are realistically many phones above 400px wide (the XS Max is 414px).

![Screen Shot 2020-06-01 at 3.05.11 PM.png](https://images.zenhubusercontent.com/5aa82c4b4b5806bc2bcab4fa/6a7b068c-4852-4b28-90c4-9f575369f043)
![Screen Shot 2020-06-01 at 3.04.58 PM.png](https://images.zenhubusercontent.com/5aa82c4b4b5806bc2bcab4fa/6169517e-6992-4545-931e-2fdc62e9c6e8)

[...]

PR #67, opened 2020-04-15
issue list page styles (#17)

thatbudakguy:
refs #17 
adds styles for the issue list page, along with a mock second issue for testing.


thatbudakguy:
I think the red arrow only showing on hover would be inconsistent with all the other places we currently have it, but will defer to @gissoo.

PR #69, opened 2020-04-16
Styles/single issue

thatbudakguy:
this PR:
- refactors the issue summary styles somewhat (the appearance on the homepage/single issue page is now the default)
- adds styles for the single issue page

it does not add styles for short pieces, since we haven't implemented those yet.


thatbudakguy:
good catch on this, thank you - the method I was using to create a white background couldn't scale. switched to a somewhat weirder but apparently effective one.

Issue #29, opened 2020-03-16
As a reader, I want to save an article using the Zotero citation manager.


thatbudakguy:
## testing notes
Go to the test articles on the test site and try harvesting with Zotero. Check that the metadata in Zotero is complete and matches the values you see on the page.

----

moved from https://github.com/Princeton-CDH/startwords/issues/7#issue-562723053

rlskoeser:
I remembered when Grant and I were reserving DOIs, I think Zotero can harvest metadata from DOIs linked on a page — so maybe we get this for free, by virtue of using DOIs.  I think DOI metadata is going to be much better than COINs .

Anyone want to test this to confirm?

rlskoeser:
I did some tests, and it doesn't seem like Zotero can reliably harvest the DOI. What's more, on the DBV piece it's picking up metadata from Sketchfab instead and that seems to supersede other things on the page.

I think it's better to go with something we know Zotero will be able to detect and handle.

thatbudakguy:
whoa, good to know. thanks for testing that out.

rlskoeser:
I think we should use meta tags, since it's more current than COinS and should be useful for both Google Scholar and Zotero (and possibly others).

Documentation here: https://scholar.google.com/intl/en/scholar/inclusion.html#indexing 

Not strictly related to this task, but I'd like to include the built in [OpenGraph template](https://gohugo.io/templates/internal#use-the-open-graph-template) and [Twitter card template](https://gohugo.io/templates/internal#twitter-cards).

(Documenting here so I can close a couple of tabs in my browser...)

gwijthoff:
Verified that everything works properly in Zotero, all metadata in its right place.

<img width="390" alt="Screen Shot 2020-06-04 at 2 49 12 PM" src="https://user-images.githubusercontent.com/5580840/84434247-aa45d400-abfd-11ea-8ce5-151a597788ca.png">

PR #79, opened 2020-06-03
Add metadata citation information for zotero/google scholar #29

rlskoeser:
Adds a new `page_meta` block to the base template and then populates it for the article page with appropriate article metadata. A few things are still yet to be determined (e.g. how we want to handle PDF urls), but I made notes so we could easily add them later.

Also adds the doi link to the article html version (no styles) and revises the text version slightly.


themes/startwords/layouts/article/single.html
```diff
@@ -14,11 +33,13 @@
         <h1>{{ .Params.Title }}</h1>
         {{ partial "authors" . }}
         {{ partial "pubdate" . }}
+        {{ if .Params.doi }}
+         <a href="http://doi.org/{{ .Params.doi }}" rel="alternate">doi:{{ .Params.doi }}</a>
```
thatbudakguy, line 37: if you throw a `class="doi"` on this link it will use the right style, which adds the `doi:` text in a `::before` so you don't have to insert it into the link text.

should we use `rel="alternate"` for the DOI links to articles on the homepage/issue page also, or does it only make sense when viewing the article itself?

rlskoeser: Cool, thanks — I thought about adding a doi class for later styling, forgot it was already meaningful with the prefix defined elsewhere.

I think `rel="alternate"` should only be used on the page for the article itself, since the relation is relative to the current url.

rlskoeser:
@gwijthoff are you able to review? I thought it would be good for you to see the metadata I'm including in the header (as well as my comments) before you test it with Zotero.

Issue #59, opened 2020-04-13
As a reader, I want to see featured content on the site's homepage.


thatbudakguy:
# testing notes
## designs
figma links have been updated to point at specific boards.
[mobile](https://www.figma.com/file/MEt3gfKX4N3WTLPbLGBbGC/Startwords?node-id=239%3A1990) / [tablet](https://www.figma.com/file/MEt3gfKX4N3WTLPbLGBbGC/Startwords?node-id=210%3A22163) / [desktop](https://www.figma.com/file/MEt3gfKX4N3WTLPbLGBbGC/Startwords?node-id=819%3A16031)

## notes

there are still a few things to keep in mind here:
- we now have designs for desktop, so please test all screens!
- the _desktop_ version of the footer isn't done yet; that will be #83.
- there is one remaining differences from the current design - I wasn't able to get the red arrows next to article previews to appear in the way that they do in the design (see [slack conversation](https://cdhprinceton.slack.com/archives/CR6JCFQHE/p1586886086062600)). They are positioned relative to the titles instead. I added the slight movement behavior on hover that @gissoo and I talked about; we didn't discuss details so feel free to suggest a different interaction (or no interaction) if it looks weird.

@gwijthoff you might want to try playing around with the preview text for the two articles (using `<!-- more -->`) to get a feel for how they will truncate. the space requirements differ slightly depending on the screen - tablet is actually the most restrictive.

gissoo:
@thatbudakguy I can't view the "homepage link" – gives me "internal error" 
Also, if this is about the homepage, could you change the figma link to show the "homepage" page on Figma? – I don't think the "Tablet" page on Figma is the appropriate one to look at

thatbudakguy:
updated the links, thanks!

gissoo:
@thatbudakguy 
below are my notes:
On all devices: 
1. The "read more" below the issue abstract is not matching the design, it needs to be left-aligned (aligned with the issue abstract) and slightly closer to the last line of the abstract than it is now. 
2. The "Read more about Startwords." needs to be replaced with a red arrow to match the design. 
3. The doi is missing, which needs to be below the authors. 
4. The length of the article previews need to be similar, this will be a problem that we need to fix. 
5. It will be better to have shape for the bottom of the preview text, however I don't think it's a deal breaker now, but would be nice to fix it for the next issue. 
6. I like the interaction of the red arrows! Thank you for doing this!

On mobile:
[...]

PR #81, opened 2020-06-04
testing article preview wrapping solution

thatbudakguy:
opening so I can get some help with this, since I'm running out of options to try and not sure how to proceed. see comments on #59 for context.

at the moment I'm trying to truncate in the template to a set number of characters and append an ellipsis if the text was truncated. the issue is that if the ellipsis comes at the end of a word that is outside the text-shape, it will not be shown at all, because the word can't get fit into the shape even though the actual text is within the character limit. `text-overflow: ellipsis` is also no help because it only works in the text flow direction - our content is overflowing the shape vertically (by one line) rather than overflowing horizontally (at the end of the line). you can see the issue in this branch; some content will show the ellipsis but on tablet (for example) it wraps outside the shape and doesn't get shown.

if there's no sane way to do this automatically, we may just have to put the burden of correctly formatting preview content on the writer, with the knowledge that they will have to manually tinker until it looks right.

### DIFF
--- a/content/issues/1/_index.md
+++ b/content/issues/1/_index.md
@@ -8,4 +8,4 @@ date: 2020-01-01
 slug: 1
 ---
 
-Data Transformations are marshmallow cotton candy candy candy canes icing toffee apple pie jujubes. Tart cotton candy danish soufflé gummies. Wafer sesame snaps cheesecake pie tootsie roll jujubes caramels jelly cake. Ice cream chocolate cake dessert. Chocolate cake cake jelly croissant pudding muffin croissant lollipop. Cookie brownie jelly-o tootsie roll dessert candy canes. Macaroon gummi bears chocolate bar ice cream gingerbread apple pie halvah.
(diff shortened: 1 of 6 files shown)


rlskoeser:
@thatbudakguy did you try changing the Hugo configuration for `summaryLength` or does that not meet your requirements for some reason (the ellipsis, maybe?)?  Default is 70 but it is configurable — see https://gohugo.io/getting-started/configuration/

This seems to me like something that should be solved by editorial practice if necessary, rather than adding complexity to the code. I'd rather have a human editing down my opening lines than a script....

thatbudakguy:
I did - see https://github.com/Princeton-CDH/startwords/issues/59#issuecomment-614882091 - but it didn't have any effect, I think because the default order of precedence places the `<!--more-->` summary above the autogenerated one and doesn't respect length.

rlskoeser:
@thatbudakguy we determined we don't need this anymore, right? Should we close?

thatbudakguy:
yep, good call. I'll close it.

PR #91, opened 2020-06-18
Feature/short pieces

rlskoeser:
- adds a draft "short piece" (one of the howtos connected to DBV essay)
- revises single issue template to treat the first two articles in an issue as the featured article, display all others in a simple list; ordering based on page metadata so editor configurable
- list of other articles is based on Gissoo's design but with a few more pieces of information (authors, tags)

I noticed that the lists I'm using in the howto don't work very well in the text output. I fiddled with it a bit but didn't find a solution.

### DIFF
--- a/themes/startwords/assets/scss/article/_summary.scss
+++ b/themes/startwords/assets/scss/article/_summary.scss
@@ -150,11 +150,16 @@ $preview-height: rem(250px);
     }
 }
 
+/* non-featured articles in an issue have a briefer listing */
+.article-short-summary {
 
-.other-articles {
     clear: both;
-    padding-top: 2rem;
     font-size: rem(16px);
+    padding-bottom: 0.75em;
+
+    &:first-of-type {
+        padding-top: 2rem;
+    }
 
     a {
         font-weight: normal;
--- a/themes/startwords/assets/scss/article/_summary.scss
+++ b/themes/startwords/assets/scss/article/_summary.scss
@@ -149,3 +149,33 @@ $preview-height: rem(250px);
         }
     }
 }
+
+
+.other-articles {
+    clear: both;
+    padding-top: 2rem;
+    font-size: rem(16px);
+
+    a {
+        font-weight: normal;
+    }
+
+    .tag {
+        color: $light-grey;
+        text-transform: uppercase;
+        font-size: rem(12px);
+        &::before {
+            content: "#";
+        }
+    }
+
+    // sequential author display with commas
+    .authors {
+        padding: 0;
+        margin: 0;
+        li, address { display: inline; }
+        address::after { content: ', '; }
+        li:last-child address::after { content: ''; }
+    }
+
+}
--- a/content/issues/1/body-as-data/index.md
+++ b/content/issues/1/body-as-data/index.md
@@ -1,7 +1,7 @@
 ---
 type: article
 title: The Body As Data
-article_number: 2
+order: 2
 authors:
     - Rebecca Munson
 date: 2020-01-01
--- a/content/issues/1/data-beyond-vision/index.md
+++ b/content/issues/1/data-beyond-vision/index.md
@@ -1,7 +1,7 @@
 ---
 type: article
 title: Data Beyond Vision
-article_number: 1
+order: 1
 authors:
   - Rebecca Sutton Koeser
   - Gissoo Doroudian
--- a/content/issues/1/stack-membership-activities/index.md
+++ b/content/issues/1/stack-membership-activities/index.md
@@ -1,0 +1,34 @@
+---
+type: article
+title: Stack Shakespeare and Company Membership Activities
+order: 6
+authors:
+  - Xinyi Li
+  - Rebecca Sutton Koeser
+tags: [howto]
+---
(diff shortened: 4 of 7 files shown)


themes/startwords/assets/scss/article/_summary.scss
```diff
@@ -149,3 +149,33 @@ $preview-height: rem(250px);
         }
     }
 }
+
+
+.other-articles {
```
thatbudakguy, line 154: probably want a comment at the top of this to explain, since it doesn't technically use the article summary template even though it is a summary of an article.

themes/startwords/assets/scss/article/_summary.scss
```diff
@@ -149,3 +149,33 @@ $preview-height: rem(250px);
...
+    a {
+        font-weight: normal;
+    }
+
+    .tag {
+        color: $light-grey;
+        text-transform: uppercase;
+        font-size: rem(12px);
+        &::before {
+            content: "#";
```
thatbudakguy, line 168: really loving the way the tags look!!

themes/startwords/assets/scss/article/_summary.scss
```diff
@@ -149,3 +149,33 @@ $preview-height: rem(250px);
         }
     }
 }
+
+
+.other-articles {
+    clear: both;
+    padding-top: 2rem;
```
thatbudakguy, line 156: i tried adding another short piece and everything looked good but there wasn't much spacing in between them on the issue list page - might want a margin-bottom here, or alternatively a rule like
```scss
.article-short-summary + .article-short-summary {
    margin-top: rem(10px);
}
```

themes/startwords/assets/scss/article/_summary.scss
thatbudakguy: it also would be a little more clear if the main heading/target for the rule was `.article-short-summary`.

rlskoeser: I wasn't sure how to do that! It seemed like it was too different from the other summary format. Would you revise the approach here, or just use the more readable target?

rlskoeser:
@gwijthoff do you have time to review this PR so we can merge and try it out on the site? I thought you should review since you'll be the one using the functionality most.

It occurred to me, I should probably add documentation to the theme readme about how this works.... I'll see if I can quickly add that so it's part of this PR and you can review.

gwijthoff:
Reproduced locally and everything works just fine, thanks @rlskoeser. I agree with @thatbudakguy that the list of short pieces could be better identified as belonging to the main "DBV" essay. Maybe a header that reads "Exercises" or "Tutorials" or something along those lines? Also, the title of the short piece is so long that it runs beneath RM's essay on the issue page, which makes it less clear that it belongs to "DBV." Can we line break halfway between the two articles? What do you think @gissoo?

Finally, what's our thinking about using or not using a date in the YAML for a short piece? I don't think it *needs* a date, I'm just curious why you didn't want to include one @rlskoeser. Also, just let me know if you'd like me to wait to merge this PR until you put documentation to the README.

Screen Shot 2020-06-23 at 3 18 40 PM

gwijthoff:
Thinking more about this and remembering that we said that 1) future Issues may include short pieces that don't belong to either of the two feature articles, and are simply related to the broader theme of the Issue, but that 2) in the case of Issue 1, the short pieces belong to one essay ("DBV") in particular.

In case that informs how we go about this here @rlskoeser @thatbudakguy @gissoo

rlskoeser:
@gwijthoff not including a date was an oversight! I think short pieces should look and work just like articles in most ways (maybe everything but how they show up on the issue page).

I didn't intend for the essays to look as if they belonged to the DBV essay — on the screens I tested it wasn't as obvious how much they lined up under it. Definitely defer to @gissoo on how they should be listed; visually, this is meant as a draft implementation based on Gissoo's existing issue list design, but adding information I wanted to see on the list view.  (I think @gissoo has a design task to revisit this — maybe you can weigh in with any goals or constraints.)

We had some discussion somewhere, but I don't remember which communication medium and if you were involved (I think you remembered our previous decision differently than I did). I _don't_ think we should explicitly link short pieces to feature articles, in either the metadata or the layout.  It's more complicated and we don't have any idea yet if it's something that's likely to happen again — and I agree with you, could easily see short pieces not explicitly related to either feature. I think we had previously agreed that we would link the DBV short pieces and essay in the text of the articles.

I did add one line to the theme readme about the article order parameter — do you think that's sufficient? I think this can be merged

gwijthoff:
Ok, so Issue 1's short pieces will all be in the text of the "DBV" article itself, this PR simply implements the architecture of how the site will handle short pieces going forward, and designing how short pieces will be displayed in future *Startwords* Issues is a separate design issue tracked elsewhere by @gissoo. Sounds good.

Just looked at the README and that's clear enough to me. Merging…

gissoo:
@gwijthoff that issue is being tracked here: https://app.zenhub.com/workspaces/startwords-5e6f8b374b152d3a86860892/issues/princeton-cdh/startwords/86

PR #95, opened 2020-06-24
Revise article text template and handle alt text + extended descriptions

rlskoeser:
@thatbudakguy @gwijthoff I have a revised approach to generating the plain text version of the article that I want you to share. It might be a little crazy... very interested to see what you two think.

After seeing how poorly the lists in the how-to short piece looked in the plain text version, in combination with trying to figure out how to display alternate text for images and other non-text content, I decided we needed to start with the markdown content first. 

I've written the article `single.txt` template to start with the `.RawContent` and use a series of regular expression replacements to clean up and revise the markdown for display as plain text. I thought it would be possible to put the regular expressions in a configuration file or data file and iteratively apply them to the content (maybe using [scratch](https://gohugo.io/functions/scratch#readout)) - that would be a lot more readable and would allow us to document the regular expressions. I worked on that approach for a while and couldn't get it to work (possibly related to string escaping — I'm not sure if the regular expressions weren't matching or I wasn't updating the scratch value correctly), so I put that aside. If we like this approach, there are still a few more regexes I should add to clean up the content.

@gwijthoff I think there are editorial implications to this approach. It gives you a lot more control over what the text version looks like, but it also puts a lot more weight on the markdown version (since in this approach that is the basis for the plain text). In terms of control: the regexes mean you can decide how you want bold, italics, links, footnote and footnote references, etc to look in the plain text. I've also added a new "txt-only" class as a simple way to provide fallback text content for images (editorial choices there too, I think — do people reading the plain text care about the images? do you want them to know what they're missing, or just give them uninterrupted access to the text?). New content might require revising/expanding the regular expressions, and you'll prolbably have to review the plain text versions more carefully. 

In addition to the revised text template, changes include:
- revised version of Hugo's built-in figure shortcode (adds option for setting `described-by` id)
- new shortcode to wrap a block and set css class and id
- new `txt-only` css class for fall back text content (It occurs to me now, the figure shortcode could create the fallback text content based on figure attributes — @gwijthoff would it make sense to have the text version automatically generated, or would you rather have more control over it?)

I've tweaked the DBV draft to help with testing, including adding some images — one of which is the concept map with a longer extended description which should be readable to screen readers and also be included in the plain text article.

Here's the current regex cleanup:
- replace markdown image with "[image: alt text]"
- remove hugo shortcodes
- remove markdown header markers (#, ##, ###)
- convert footnote label [^1]: to 1. 
- convert markdown links to "link text [URL: link]"
- convert bold text to regular text (`**word**` becomes `word`)

Not yet handled: 
- footnote references in text (wasn't sure what they should look like)
- anchor links within the document (likely remove)
- anchor names on headers (remove)
- italics or any other formatting
- pull quotes (ignore? or add custom ascii art and include?!?)


gwijthoff:
@rlskoeser Working through this PR, but for now, I wonder if this part is something you can share as an issue on the [Hugo GitHub repo](https://github.com/gohugoio/hugo/issues)? Seems like something a lot of people would be interested in.

> I thought it would be possible to put the regular expressions in a configuration file or data file and iteratively apply them to the content (maybe using scratch) - that would be a lot more readable and would allow us to document the regular expressions. I worked on that approach for a while and couldn't get it to work (possibly related to string escaping — I'm not sure if the regular expressions weren't matching or I wasn't updating the scratch value correctly), so I put that aside. If we like this approach, there are still a few more regexes I should add to clean up the content.

rlskoeser:
@gwijthoff I was wondering about asking on the [discussion forum](https://discourse.gohugo.io/) about customized text output! I was looking around for other options for how to approach this — it seemed plausible to me that one of the markdown renderers could have a text output option, but I don't think any such renderer exists  — and if it did, I'm not sure how to tell hugo to use it for the text template.  Would you ask about using regular expressions specifically or ask the broader question about best approach for customized text output?

gwijthoff:
@rlskoeser regex is your proposed solution to the broader problem of overly blunt, not-well-formatted text output, right? Seems like you could describe the issue, then say "I got this far but no farther, any ideas?" Or something along those lines.

gwijthoff:
> It gives you a lot more control over what the text version looks like, but it also puts a lot more weight on the markdown version (since in this approach that is the basis for the plain text).

You have no idea how much I'm geeking out over this right now.

rlskoeser:
> You have no idea how much I'm geeking out over this right now, @rlskoeser

@gwijthoff I thought you might appreciate it!

rlskoeser:
Ooh, possibly relevant: https://discourse.gohugo.io/t/output-text-file-with-fixed-line-width/8588

gwijthoff:
@rlskoeser omg @thatbudakguy and I *just* had to hop on the phone to talk through soft/hard line wrapping issues. I'll spare you the trip down the rabbit hole we found ourselves in, but the place we settled is that hard wrapping (whether in a markdown file or in the .txt output) is, in effect, an authorial or editorial decision made without the input of the author. So, it's almost a kind of textual criticism argument that best practice here would be to never hard wrap or break lines in places the author didn't intend. Even if that means the .txt file in a desktop browser is comically long. For me, the main affordances of that .txt file are sustainability (it will always be human- and machine-readable as long as humans and machines are around) and reproducibility (people can take that file and interpret, remix, re-typeset it in new ways).

gwijthoff:
@rlskoeser re: image captions in plain text, I'm leaning toward having the .txt version auto-generated from the figure shortcode, rather than having a separate txt-only css class. In most cases, it seems like the image caption used for the HTML and PDF versions will work perfectly in TXT. Having two separate markup structures for the same image might prove confusing.

But one thing we might want to consider in addition to using `caption` and `attr` fields in the figure shortcode (which are both currently there) is using an `alt` field for a textual description of the image. I'm no expert on web accessibility, so @thatbudakguy might have thoughts here, but we should have alternate image descriptions anyway, right? And those alt descriptions could be included alongside the `caption` and `attr` fields in the .TXT output.

I just tried doing this, however, and it looks like the figure shortcode as written translates the `caption` field into `alt` for HTML. Can we have three separate fields in the shortcode, `caption`, `attr`, and `alt` for the fullest possible remediation of the image into the .TXT format? Or is that confusing things?

gwijthoff:
@rlskoeser where exactly are these regex transformations taking place? As part of the Hugo build? Where can I see / change the regex?

rlskoeser:
@gwijthoff yes, I thought that would probably make sense (after I implemented it the first way it seemed like it would be redundant). Gad you agree! Do you want to suggest how you'd want figures to look in text format? You'll see that for markdown images I'm using the alt text also.

We should definitely use `alt` tags on images! My oversight for not including any in the preliminary figures I added. The hugo figure tag definitely supports it.

themes/startwords/layouts/article/single.txt
```diff
@@ -28,6 +28,8 @@
 |  {{ .Parent.Date.Format "January 2006" }}
 |
 +----------------------------------------------------------------------------------->
+{{/* remove all tags; convert footnote list; convert markdown URLs and images; remove heading marker.
+    (could we configure in site config or data file and apply iteratively?)*/}}
+
+{{ .RawContent | replaceRE "!\\[([^\\]]+)\\]\\([^\\)]+\\)" "[image: $1]" | replaceRE "{{\\<[^\\>]+>}}" "" | replaceRE "(?m)^#+ " "" | replaceRE "(?m)(<\\/?[^>]+>)" "" | replaceRE "(?m)^\\[\\^(\\d+)\\]: " "$1. " | replaceRE "\\[([^\\]]+)\\]\\(([^\\)]+)\\)" "$1 [URL: $2]" | replaceRE "\\*\\*([^*]+)\\*\\*" "$1"  }}
```
rlskoeser, line 34: @gwijthoff here's where the regexes are currently! in the single article text template

gwijthoff: Are we sure this isn't ASCII art? Always amazes me when first looking at them that regexes are actually functional! 😀

I don't have enough knowledge to do much with this, but can always ask @rlskoeser if I notice any issues during the editing process. For now, the transformations on the other side look brilliant, and I'm so excited to have such a beautiful and functional .TXT file.

The one thing I can think of is using a `====` or `-----` underline for section headings, rather than `#` or `##`, respectively. Both pandoc and github-flavored markdown accept that underlining format for section headings, and they also have the benefit of being pretty striking to a reader's eyes on the page.

So, is there a way to edit the above regex in this `single.txt` file to have a section heading display as, for example:

```
Weaving Derrida's References
============================
```

I suppose the regex would have to count the number of characters in the header text and then print the same number of `=`'s. 28 characters in that header text above, and 28 `=`'s beneath it.

rlskoeser: The regexes are a lot less readable than usual because I had to use double slashes to escape the slashes (took me a bit to figure out). I figured out at some point that I could use ` instead of " in which case I didn't seem to need the doubles, but I was worried about changing it and breaking working regexes!

I like your idea about the alternate header format! It looks like common mark / goldmark supports this syntax too — if we just use that format in the original markdown then we don't have to transform it at all for the plain text version. I can test this.

rlskoeser:
@gwijthoff ok, pushed an update to use setext headings (which is apparently what they are called) for the h2s in the DBV essay. Unfortunately setext only supports h1 and h2.   IDK if it's weird to use a mixture of setext and atx headings (seems fine to me?).

See what you think of how these look in the text version.  I like the idea of not having to clean them up with regexes.

gwijthoff:
@rlskoeser Coming back to the question of figures in TXT. An example might be, shortcode:

```
{{< figure src="images/conceptmap.jpg" caption="Concept map situating data physicalization in relation to other types of data representations and interpretations." attr="From the “Data Beyond Vision” poster presented at DH2019, Utrecht, July 2019." alt="Purple text on white background with lines joining forms of data transformations to their sensory components." attrlink="https://doi.org/10.5281/zenodo.3261531" desc_id="conceptmap-desc">}}
```

and a maybe slightly creative .TXT version that would include 1) a markdown formatted version of the image so that it's typeset-able, and 2) human-readable captions:

```
<!--IMAGE-->
![http://localhost:1313/issues/1/data-beyond-vision/images/conceptmap.jpg](Concept map situating data physicalization in relation to other types of data representations and interpretations.)
<!--
DESCRIPTION: Purple text on white background with lines joining forms of data transformations to their sensory components.
CAPTION: Concept map situating data physicalization in relation to other types of data representations and interpretations.
ATTRIBUTION: From the “Data Beyond Vision” poster presented at DH2019, Utrecht, July 2019.
LINK: https://doi.org/10.5281/zenodo.3261531
-->
```

I guess I'm starting to think now about the ways this .TXT file will or won't be pandoc-friendly. That's why I'm thinking in terms of commenting out the figure descriptions. It might be a little ugly though. What do you think?

rlskoeser:
> I guess I'm starting to think now about the ways this .TXT file will or won't be pandoc-friendly. That's why I'm thinking in terms of commenting out the figure descriptions. It might be a little ugly though. What do you think?

Interesting, can you say more about this use case?   I wondered why you were using comments. It does seem less readable...

gwijthoff:
Setext!!

And wow that looks nice! Thanks for the fast work! Out of curiosity, why would `Folding in the Non-Famous Members` be H2 and not H1? Why not make `Goal` be H2?

Screen Shot 2020-06-25 at 3 45 05 PM

rlskoeser:
Because the title of the whole article is an h1 and you can only have one h1 (unless there's a way to make these sections so they can have nested h1s... — although then we'll have to update our styles).

gwijthoff:
> Interesting, can you say more about this use case? I wondered why you were using comments. It does seem less readable...

It comes back to this:

> For me, the main affordances of that .txt file are **sustainability** (it will always be human- and machine-readable as long as humans and machines are around) and **reproducibility** (people can take that file and interpret, remix, re-typeset it in new ways).

A use case highlighting the second affordance would be (for whatever reason) someone wanting to take the articles from our site (or any site using this Hugo theme) and transform them into another format using Pandoc or any other kind of digital typesetting. Making .TXT with markdown available is close enough to that goal anyway, and I suppose there's no reason we necessarily *have* to produce .TXT files that could instantly be used by Pandoc. We can't anticipate all of the ways people will want to transform that .TXT file. 

So I may have just convinced myself that something more human-readable would be better for the .TXT files, as opposed to markdown that's computer-readable. It is a .txt and not an .md file, after all. But I'm curious to hear your thoughts.

A more human-readable version might look like:

```
+----------------------------------------------------------------------------------->
| [IMAGE]
|
| DESCRIPTION: Purple text on white background with lines joining forms of data transformations to their sensory components.
| CAPTION: Concept map situating data physicalization in relation to other types of data representations and interpretations.
| ATTRIBUTION: From the “Data Beyond Vision” poster presented at DH2019, Utrecht, July 2019.
| LINK: http://localhost:1313/issues/1/data-beyond-vision/images/conceptmap.jpg
+----------------------------------------------------------------------------------->
```

Or something fun like that.

rlskoeser:
I like that revised image format better — at least, it's much more appealing to me visually. It seems like it wouldn't be as machine readable, FWIW; if we want it to be parsable, something more like this might be easier to work with (thinking of contention of`#` as commented lines):
```
#----------------------------------------------------------------------------------->
[IMAGE]

DESCRIPTION: Purple text on white background with lines joining forms of data transformations to their sensory components.
CAPTION: Concept map situating data physicalization in relation to other types of data representations and interpretations.
ATTRIBUTION: From the “Data Beyond Vision” poster presented at DH2019, Utrecht, July 2019.
LINK: http://localhost:1313/issues/1/data-beyond-vision/images/conceptmap.jpg
#----------------------------------------------------------------------------------->
```

Side note: do we need to distinguish between figures and images? maybe try to use figures as much as possible?

For the remixing... our markdown files will be available too, on github at least, although I don't think we'd want to distribute them on the public site the way we do the other contents...

gwijthoff:
@rlskoeser what's the figure / image distinction on the site? Images in contextual notes, and figures are numbered in the body of the article text? Or?

rlskoeser:
@gwijthoff we can use markdown images if we want to, and I have a simple regex now that converts it to label as an image and show the alternate text. Just thinking it might be a good policy to stick with figures as much as possible, though, for the nicer display and consistency. Not sure about endnotes, good question...  If we're using both we might want some consistency between how they show in the TXT version.

rlskoeser:
@gwijthoff do you want to request changes on this PR and document what you'd like to see changed before we merge it? My sense was that included:
- revising plain text version of figures (include it in the figure snippet, add text markers of some kind around it)
- additional regexes to clean up a few more things I'm currently missing (anchor links at least)

I should probably add some notes to the theme documentation about how the plain text output works and what conventions we're using for markdown — agree?

Do you want a plain text version of pull quotes analogous to the figure, or should pull quotes be skipped in the plain text?

I'll likely ask a question about this on the Hugo support form sometime, but I think it might be easier to refer to the draft site so they can see how it looks as well as seeing the code.

gwijthoff:
> @gwijthoff do you want to request changes on this PR and document what you'd like to see changed before we merge it? My sense was that included:

Sure, doing that now

> I should probably add some notes to the theme documentation about how the plain text output works and what conventions we're using for markdown — agree?

Sounds great!

> Do you want a plain text version of pull quotes analogous to the figure, or should pull quotes be skipped in the plain text?

I think we can forego the pull quotes in plain text.

rlskoeser:
@gwijthoff have updated this based on what we decided in our previous conversations.

The one thing I couldn't do was have the figure short code generate the plain text version of the figure — I had forgotten that the _results_ of the short code template aren't included in the content we're working from to generate the plain text, since I'm using the raw markdown as the starting point.  

I think it might be possible to use a regex to construct the plain text figure from the figure short code, but I'm not sure how reliable it would be, and I'd rather not implement it now. I've added documentation to the readme as we discussed, including a template for plain text figures.  (For simplicity, I adjusted slightly to avoid using characters that get converted into entities in the text output.)

gwijthoff:
Love the way the image captions look in plain text! My one very minor thought is whether we should use `+` characters at the corners of the image caption boxes, rather than `#` characters in case they throw off any markdown processing? The box at the top of the document's metadata header uses `+`'s. And, do you have thoughts on using a `>` at the end of the line, like in the header boxes, so that they're all uniform and match the *Startwords* visual identity? So, 

```
+----------------------------------------------------------------------------------->
|
|
|
```

Rather than

```
#-----------------------------------------------------------------------------------
|
|
|
```

gwijthoff:
@rlskoeser also: I understand the issue with auto generating plain text captions in the format we want from the shortcodes, and think that the manual description of how to add the `{{< wrap class="txt-only" >}}` shortcode manually is a fine workaround for the time being.

rlskoeser:
@gwijthoff I tried to say something about this but I recognize now it wasn't very clear or specific at all!
> For simplicity, I adjusted slightly to avoid using characters that get converted into entities in the text output.

What I meant is — when I put `+` and `>` in the markdown, they get converted to `&#43;` and `&gt;` respectively. There might be a way to clean them up, but the things I tried didn't work (adding `htmlUnescape` didn't help at all) and I'd prefer not to add more regular expressions if we can avoid it.

I agree that it would be better if the characters used to mark the header and figures matched. Would you be open to revising the heading and finding characters that don't get converted like this?

gwijthoff:
@rlskoeser found some fancy math symbols in unicode that work better for forming figure and header lines. So, looks like we're good to go on this one!

PR #97, opened 2020-06-24
Basic postcss + autoprefixer setup

rlskoeser:
Setup basic postcss + autoprefixer based on the hugo docs. Working on updating the github actions for the change.

The readme (when we have one, or developer notes) might need to indicate that `npm install` needs to be run now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,24 @@
+{
+  "name": "startwords",
+  "version": "1.0.0",
+  "author": "The Center for Digital Humanities at Princeton <cdhdevteam@princeton.edu>",
+  "description": "",
+  "license": "Apache-2.0",
+  "private": true,
+  "dependencies": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "startwords",
+  "version": "1.0.0",
+  "author": "The Center for Digital Humanities at Princeton <cdhdevteam@princeton.edu>",
+  "description": "",
+  "license": "Apache-2.0",
+  "private": true,
```
thatbudakguy, line 7: since `private = true` and we'll never publish the package, you're free to take out the generated `repository`, `scripts`, `bugs`, `homepage`, `description`, etc. keys if you want. I think npm/yarn only check for the presence of `dependencies` and `devDependencies`, though they might warn about other things missing.

PR #110, opened 2020-07-23
adding a flourished section break

gwijthoff:
Had another one of those "can't believe we haven't thought of this before" moments this morning. I don't think we've implemented anything for section breaks yet. @gissoo how do you feel about using the logo for a [section break](https://en.wikipedia.org/wiki/Section_(typography))? @thatbudakguy what do you think of how this is implemented?

Screen Shot 2020-07-23 at 8 15 54 AM


rlskoeser:
@gwijthoff I know you didn't tag me, but couldn't help seeing this and thinking about it. I like the idea, but it would be even better if we could use a semantic section break (I imagine this would be better for screen readers). I think we might be able to use a decorative section break in markdown ([some examples](https://docs.moodle.org/39/en/Advanced_use_of_Markdown) — I presume that renders as a horizontal rule when the markdown is converted to html, and then we can use CSS to style it however we want (meaning we could make it look the same as what you have implemented currently). This has an additional benefit, because the section break then will also be visible in the text output.

I'm interested to hear what @gissoo thinks in terms of design, when she has a chance to think about it.

gwijthoff:
@rlskoeser great idea! I like giving authors more control over and flexibility with the markdown, too. Maybe we can have both `***` and `* * *` function as section breaks.

gissoo:
@gwijthoff @rlskoeser Thank you for writing!! I like this! But could you clarify? What does a "section break" mean in the context of an article? How are we defining a "section" in an article? 
Can it be between every paragraph? Or only above a paragraph that has a different title from the paragraph above it?

thatbudakguy:
agree with @rlskoeser's comment that it would be best to do this as an `<hr/>` so we get the semantics too. if i'm reading those examples right, `***` and `* * *` will both work fine without any effort on our part, since the parser is just looking for (at least) three consecutive asterisks, with any amount of spaces allowed.

@gwijthoff if you wanted to play around with it, you could change the css in this PR to target the `hr` element instead of `.sectionbreak`, and then maybe use the `background-image` property to add the image via css (instead of an image tag). i can take a crack at this as well.

@gissoo my guess is that a section in this context is probably several paragraphs, and it likely is followed by several more paragraphs. if you think about long essays, the writer often doesn't use headings to break up the text, but there are groups of paragraphs that talk about the same content or are related in some way, so it's helpful to visually separate them even though there's no heading. i think this section break might be intended for that use case (but just a guess - @gwijthoff could explain more).

gwijthoff:
@gissoo good questions! The section break character is typically used in essays, novels, and stories when there is no section heading. I've always liked this line from Kathy Acker describing the units of writing:

> Well, a comma’s a breath, and a sentence is a thought, and a paragraph is an emotion… You’re always working the paragraph against the sentence.

Not sure where that puts the section break, but to me it always feels like a pause or a transition between topics. Because DBV contains section headings/titles, that piece likely won't use the section break character. But RM's essay doesn't have section headings/titles, so will need the character.

@thatbudakguy thanks for the guidance! I can take a crack at this approach later in the week.

gissoo:
@gwijthoff @thatbudakguy thanks for writing. This aligns with my thoughts as well. Let's do this only when there are no headings/titles within a piece, let me know what you think. I'm happy with using the logo, and I'm glad @rlskoeser has thought of a better way to implement it.

rlskoeser:
@gwijthoff do you care about making the changes to the css yourself or are you busy enough that it would help to have someone else do it?

And question for everyone: do we expect to need regular horizontal rules anywhere or are we fine with them *always* being treated as a section break and styled the way Grant proposed? (i.e. is this style scoped to articles only or can it be anywhere on the site?)

thatbudakguy:
I'm ok with hrs always being the styled version! I think they serve the same purpose.

thatbudakguy:
ok, I've implemented this using `<hr>`s as discussed. I misinterpreted RSK's original question:

> is this style scoped to articles only

i think it makes sense for now to keep it scoped to articles, and it was easy to do - not sure if we'll use an `<hr>` outside of one.

one thing to note is the vertical margins for the section break currently aren't dependent on screen size - not sure if we want a more relaxed look on desktop vs. on mobile.

thatbudakguy:
@gwijthoff if you have time, pull the changes down and give it a look to see if it still looks right.

rlskoeser:
@thatbudakguy I just reviewed, it looks great to me. I think it's ok to go ahead and merge since Grant is on vacation. (He can always review on the site later.)

thatbudakguy:
oops, totally forgot. thanks for reviewing!

PR #115, opened 2020-08-21
Get replaceRE args from data file

jmooring:
@rlskoeser 

See https://discourse.gohugo.io/t/27707


rlskoeser:
Thanks again @jmooring. Have merged your solution and was able to add comments to document all of my regexes. This will be much more manageable and maintainable now.

Issue #121, opened 2020-09-10
Integrate ISSN into site config and article metadata


gwijthoff:
The week of September 21, I will apply for an ISSN for the journal from the Library of Congress. [See Asana task here](https://app.asana.com/0/1152092590277803/1192371786575251). The turnaround time is usually two weeks. Once we have it, we should add the ISSN into the site's footer and metadata.

rlskoeser:
I think I misremembered the guidelines! For online publications it only needs to be displayed on the home page or main menu:

https://www.issn.org/understanding-the-issn/assignment-rules/the-issn-for-electronic-media/
> How should an ISSN be displayed on electronic publications?
> Online publications (websites, blogs, wikis, databases, etc.)
> The ISSN should be shown on the homepage or, failing that, on the main menu.

And for online journals:
> For serials available only in online versions, the ISSN should appear on the title screen or home page and/or in the masthead or other areas where information about publisher, frequency, subscribing, copyright, etc. is given.

This seems better than revising the footer to try to fit it in. @gwijthoff do you agree? Are you comfortable adding it to the markdown yourself or do we want it somewhere else on the home or about page generated from the site config file?

gwijthoff:
@rlskoeser I think adding the ISSN to the About page would work best. Seems like it would clutter our beautifully designed homepage. I can add it to the markdown myself, no problem!

What about the site config / metadata? Any reason the ISSN should be added there, whether for Zotero harvesting or Google visibility, etc?

rlskoeser:
@gwijthoff I just looked to confirm, I already have a placeholder! https://github.com/Princeton-CDH/startwords/blob/master/themes/startwords/layouts/article/single.html#L23

I think we should add it to the site config and then use that to populate this field in the article citation metadata (which is used by both Zotero and Google Scholar).

gwijthoff:
@rlskoeser of course you do! Okay, I'm taking Nick off this task and assigning myself. I can add the About page markdown if you want to tackle the site config etc.

gwijthoff:
Cf. the [site config commit](https://github.com/Princeton-CDH/startwords/compare/b56c5478f219...08ce2437c1a8) and the [about page commit](https://github.com/Princeton-CDH/startwords/commit/00125b2243316c16837898a9eed5d048a9b27205).

PR #123, opened 2020-09-16
Add ISSN placeholder to site config; include in article metadata if set

rlskoeser:
ref #121

This is a pretty trivial change (I see now that I should have done it this way from the start!), but I thought you both would want to be aware of it and thought you could use the PR as a chance to review and — if you want — confirm on your local instance that setting an ISSN in the config does the right thing.


gwijthoff (approved):
Checks out for me! Glad this was so simple.

thatbudakguy (approved):
lgtm; sorry it took me so long to look at this!

PR #128, opened 2020-09-22
a11y and performance updates

### LINKED ISSUE #120: issue titles have insufficient contrast
**Describe the bug**
The issue titles on both the issue list page and single issue page have a contrast ratio with the background of **1.82**, which fails [WCAG AA standards](https://www.w3.org/TR/WCAG21/#contrast-minimum) for large text (minimum **3.0**).

**To reproduce**
See [this contrast ratio calculator](https://contrast-ratio.com/#rgb%28128%2C%20128%2C%20128%2C%200.5%29-on-white) that takes into account `rgba()` colors with transparency. Other contrast calculators may incorrectly ignore the fact that the color of the titles is transparent - `rgba(128, 128, 128, 0.5)`.

**Expected behavior**
All text on the site should pass WCAG AA requirements.

**Screenshots**
taken using firefox's accessibility panel in the developer tools.
Screen Shot 2020-09-01 at 5 11 30 PM

**Additional context**
The non-transparent version of the color - `rgb(128, 128, 128)` - _does_ pass WCAG requirements, with a contrast ratio of about **3.9**.

thatbudakguy:
this PR makes a number of performance and accessibility-related changes to improve lighthouse scores. there is still a lot of work to do on the main article page (at least for DBV), but this is a start.

## performance
- uses `woff` and `woff2` versions of web fonts, instead of `ttf` or the [deprecated "bulletproof" font stack](https://pixelambacht.nl/2019/youve-gotten-old-bulletproof-font-face-syntax/)
- uses [`font-display: swap;`](https://web.dev/font-display/) to avoid missing text while fonts load
- preloads critical resources, such as fonts
- [preconnects](https://web.dev/uses-rel-preconnect/) to some 3rd-party domains like sketchfab if the relevant shortcodes (e.g. deep zoom) are loaded
- uses `script async` to prevent contextual notes js from blocking the main thread
- uses `script defer` to prevent openseadragon from blocking the main thread, and loads the viewer after the page has loaded
- uses [lazy loading iframes](https://web.dev/iframe-lazy-loading/) for the sketchfab viewer
## a11y
- fixes #120 by adding a new grey color with more contrast


themes/startwords/assets/scss/_fonts.scss
```diff
@@ -1,127 +1,182 @@
 /* font familes */
 
+/* source-sans-pro-200 - latin */
 @font-face {
     font-family: 'Source Sans Pro';
     font-style: normal;
     font-weight: 200;
-    src: url('/fonts/Source_Sans_Pro/SourceSansPro-ExtraLight.ttf');
+	font-display: swap;
+    src: url('/fonts/Source_Sans_Pro/source-sans-pro-v13-latin-200.woff2') format('woff2'),
```
rlskoeser, line 9: I didn't know we were using local copies of the font, is there a reason for that? Would it be more efficient to load them from google fonts as a CDN?

themes/startwords/assets/scss/_fonts.scss
```diff
@@ -1,127 +1,182 @@
...
+         url('/fonts/IBM_Plex_Serif/ibm-plex-serif-v8-latin-700.woff') format('woff');
 }
 
+/* ibm-plex-serif-700italic - latin */
 @font-face {
     font-family: 'IBM Plex Serif';
     font-style: italic;
     font-weight: 700;
-    src: url('/fonts/IBM_Plex_Serif/IBMPlexSerif-BoldItalic.ttf');
+    font-display: swap;
```
rlskoeser, line 178: Glad to learn about `font-display : swap`, have definitely seen the FOIT behavior on my phone on other sites.

Seems like the font declarations are very repetitive, wonder if it's worth a mixin? Although maybe not worth it now.

themes/startwords/layouts/shortcodes/deepzoom.html
```diff
@@ -3,14 +3,16 @@
...
-    id: "{{ $zoom_id }}",
-    prefixUrl: "https://cdn.jsdelivr.net/npm/openseadragon@2.4/build/openseadragon/images/",
-    preserveViewport: true,
-    visibilityRatio:    1,
-    minZoomLevel:       1,
-    defaultZoomLevel:   1,
-    gestureSettingsMouse: { scrollToZoom: false },
-    tileSources: "{{ if .IsNamedParams }}{{ .Get "tile" }}{{ else }}{{ .Get 0 }}{{ end }}",
-});
+    window.addEventListener("onload", function() {
```
rlskoeser, line 6: Whoops, I should have done it that way when I implemented it!

themes/startwords/assets/scss/_fonts.scss
thatbudakguy: The main reason is that CSS is render-blocking, and google fonts are normally implemented by requesting a stylesheet which then itself connects to a different google domain to get the font ([analysis in this article](https://www.tunetheweb.com/blog/should-you-self-host-google-fonts/)). In some self-hosting situations (maybe our production VMs, haven't tested) it _could_ be a closer call since google's CDN is fast, but here we're essentially using GitHub as a CDN, so that's not an issue.

themes/startwords/assets/scss/_fonts.scss
thatbudakguy: Oh, that's a fantastic idea! I think we might use something very similar to this on other projects going forward, so I'd say it could be worth it. I'll experiment.

themes/startwords/layouts/shortcodes/deepzoom.html
thatbudakguy: OpenSeadragon isn't actually a very heavy lift; it's mainly the Sketchfab embed that is a huge resource (I think it's an entire React app with Redux!). I've tried some things to mitigate it but the lighthouse scores are still pretty bad on pages that include it, so I need to investigate more.

themes/startwords/assets/scss/_fonts.scss
rlskoeser: Got it. Thanks for explaining and providing a link.

themes/startwords/assets/scss/_fonts.scss
thatbudakguy: On revisiting this, I'm not sure the best way to do it, since the things that change each `@font-face` declaration are the style, weight, and src and all are dependent in less-than-obvious ways on the _name_ of the font file (and which weights/styles we chose to include). There might be a way to standardize it, but it requires more thinking - maybe we can revisit for Geniza.

PR #130, opened 2020-09-24
Adjust masthead formatting & content on the about page

rlskoeser:
Made a few revisions to the masthead on the about page:
- wrap the masthead so styles can be keyed to this section only
- revise the wrap shortcode to interpret markdown in wrapped content
- remove the | in the page text and add a border in the css so it looks similar but is not read by screen readers
- remove 'developer' role and make a 'technical' list with names similar to the advisors list (I think this better encompasses the range of work we did, which was not solely development)

Questions:
- @thatbudakguy better place to put the styles, to organize it more specifically?
- @thatbudakguy are you ok with 'technical' instead of developer? 
- @gwijthoff should you be listed under technical? You've worked on the site too. I don't know if it's assumed that editors are involved in that kind of work, it seems like in a lot of cases they wouldn't be
- @gwijthoff What is the logic for the order of names on the board of advisors list? Should it be alphabetical?

### DIFF
--- a/content/about.md
+++ b/content/about.md
@@ -13,19 +13,22 @@ Each issue features two articles that speak to a common theme, feature, or conce
 
 *Startwords* is built with the Hugo static site generator using a custom-made theme. In support of open access principles, we publish all articles under a Creative Commons license (CC BY 4.0) and the [website code](https://github.com/Princeton-CDH/startwords) under an Apache 2.0 license.
 
-## Masthead
+{{< wrap class="masthead" >}}
 
-**Editor** | Grant Wythoff
+## Masthead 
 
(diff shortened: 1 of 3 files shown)


gwijthoff:
@rlskoeser to answer your questions, sure! I'd be happy to be listed under technical. And you're right, the board of advisors should be alphabetical. Just added those changes.

gwijthoff (approved):

thatbudakguy (approved):

thatbudakguy:
@rlskoeser the place you picked for the styles is exactly what I'd do, so looks good to me! and I'm good with "technical" too.

PR #135, opened 2020-09-30
Basic robots.txt with sitemap url

rlskoeser:


config.toml
```diff
@@ -4,10 +4,14 @@ theme = "startwords"
...
 enableEmoji = true
 googleAnalytics = "UA-87887700-8"
+enableRobotsTXT = true
+enableGitInfo = true
 
-[taxonomies]
-  author = "authors"
-  tag = "tags"
+# disable taxonomies until we are ready to use them and have templates
+disableKinds = ["taxonomy", "term", "tag"]
```
thatbudakguy, line 11: nice, maybe hugo's builds won't emit warnings about this now.

config.toml
```diff
@@ -4,10 +4,14 @@ theme = "startwords"
 copyright = "This work is licensed under a Creative Commons Attribution 4.0 International License."
 enableEmoji = true
 googleAnalytics = "UA-87887700-8"
+enableRobotsTXT = true
+enableGitInfo = true
```
thatbudakguy, line 8: what's the `GitInfo` do?

thatbudakguy (approved):
looks good

config.toml
rlskoeser: It uses git commit info for generating last modified dates for content — see https://gohugo.io/variables/git/

rlskoeser: Should we add a comment in the config? 

From what I read this makes the build a little slower, but seems worth it to me. The github actions build is based on a git checkout, so I think it should work.

thatbudakguy: cool! seems worth it to me too. a comment wouldn't hurt.

PR #136, opened 2020-09-30
Tweak rss feed, add atom feed

rlskoeser:
- Tweaked rss feed link in header based on hugo docs so mimetype is escaped properly
- Adds a new default rss feed based on the hugo built in with slight adjustments (added author, although the spec doesn't actually handle multiple names and expects an email 🙄 )
- Add configuration and template for an atom feed — atom handles multiple authors and I found an existing project that was close to what we needed. Note that I'm not using the theme component directly because it didn't actually render any pages when I tried it; I combined the logic from hugo's built in rss template and the atom feed template.  We might need some cleanup for content we include, we might need to strip out scrip tags; but will be easier to test in some viewers once this is in place.

There's room for more customization, but I thought this would be a good first pass.

RSS and ATOM links should be in the header metadata, but they're at `index.xml` and `atom.xml` respectively.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -17,6 +17,9 @@ googleAnalytics = "UA-87887700-8"
     mediaType = "text/plain"
 
 [outputs]
+  home = ["HTML", "RSS", "ATOM"]
+  section = ["HTML", "RSS", "ATOM"]
+#  taxonomy = ["HTML", "RSS", "ATOM"]
   page = ["HTML", "TXT"]
 
 [menu]
@@ -24,6 +27,10 @@ googleAnalytics = "UA-87887700-8"
     name = "About"
     url = "about/"
     weight = 20
+
+[Author]
+  name = "Center for Digital Humanities at Princeton"
+  email = "cdh-info@princeton.edu"
 
 [params]
   dateform        = "January 2, 2006"
--- a/themes/startwords/layouts/_default/baseof.html
+++ b/themes/startwords/layouts/_default/baseof.html
@@ -60,7 +60,7 @@
 
     <!-- alternative formats -->
     {{ range .AlternativeOutputFormats -}}
-    <link rel="{{ .Rel }}" type="{{ .MediaType.Type }}" href="{{ .Permalink | safeURL }}">
+    {{ printf `<link rel="%s" type="%s" href="%s" />` .Rel .MediaType.Type .Permalink | safeHTML }}
     {{ end -}}
 
     {{ block "page_meta" . }}{{ end }}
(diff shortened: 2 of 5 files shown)


config.toml
```diff
@@ -17,6 +17,9 @@ googleAnalytics = "UA-87887700-8"
     mediaType = "text/plain"
 
 [outputs]
+  home = ["HTML", "RSS", "ATOM"]
+  section = ["HTML", "RSS", "ATOM"]
+#  taxonomy = ["HTML", "RSS", "ATOM"]
   page = ["HTML", "TXT"]
```
thatbudakguy, line 23: should this include "PDF"?

thatbudakguy: oh, no it shouldn't - we do that manually. sorry for being confused

themes/startwords/layouts/_default/baseof.html
```diff
@@ -60,7 +60,7 @@
 
     <!-- alternative formats -->
     {{ range .AlternativeOutputFormats -}}
-    <link rel="{{ .Rel }}" type="{{ .MediaType.Type }}" href="{{ .Permalink | safeURL }}">
+    {{ printf `<link rel="%s" type="%s" href="%s" />` .Rel .MediaType.Type .Permalink | safeHTML }}
```
thatbudakguy, line 63: nice, I was hoping to do something cool like this! `.MediaType.Type` is awesome.

thatbudakguy: oh wait, we can't do this for articles - can we?

thatbudakguy (approved):
LGTM; the hugo template stuff happening at the top of the atom and rss feeds is wayyy beyond my pay grade but seems like it works right!

themes/startwords/layouts/_default/baseof.html
rlskoeser: @thatbudakguy not sure what you mean about articles. This is the base template so we should get this behavior there too. I would expect it to pick up the text but not the pdf (since it isn't an alternate that Hugo knows about / manages).

thatbudakguy: sorry, I'm mixing up `<head>` metadata with the actual output format download links in the article body. was mostly wondering if there's anything going on here that we could use in the same way to generate the output format links in a "smarter" way (autodetect MIME type?), but I don't think we can.

rlskoeser: @thatbudakguy we're already using output formats similar to this to get the txt link for the article body. I think we could get the mime type from the object instead of hard coding it, but I'm not sure how much that gets us. (It would make more sense if we could do that for both pdf and text.)

PR #144, opened 2020-10-13
Override focus outline with round border for contextual note close link

rlskoeser:
The square focus outline on the round contextual note close link was bothering me so I took a few minutes to override the outline with a rounded border for that element only. Hopefully this doesn't introduce any problems.

### DIFF
--- a/themes/startwords/assets/scss/article/_notes.scss
+++ b/themes/startwords/assets/scss/article/_notes.scss
@@ -124,8 +124,8 @@ body.article .footnotes {
         }
 
         a {
-            @include highlight-focus;
-            color: $light-purple;
+             @include highlight-focus;
+             color: $light-purple;
             &:focus { outline-color: $off-white; }
         }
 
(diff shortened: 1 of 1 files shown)


themes/startwords/assets/scss/article/_notes.scss
```diff
@@ -124,8 +124,8 @@ body.article .footnotes {
         }
 
         a {
-            @include highlight-focus;
-            color: $light-purple;
+             @include highlight-focus;
+             color: $light-purple;
```
thatbudakguy, line 128: what's up with this added whitespace?

thatbudakguy (approved):
tested out locally; the dashed border gets a little warped in places but it's definitely better than just a square. no problems i can foresee.

themes/startwords/assets/scss/article/_notes.scss
rlskoeser: I don't know! I removed these and then added them back because I figured out I was changing the styles in the wrong place (not specific enough). I can check and fix.

rlskoeser:
@thatbudakguy where is it getting warped? Should I adjust the border radius? I just tried a couple of values and found one that looked decent (not very systematic at all 🙂 )

thatbudakguy:
sorry, not very well-worded. i don't think it can be helped, it's more of a css layout bug/artifact. you can see it at the midpoints of each side of the circle; the dash is sort of broken in the middle. might also be FF only; didn't test on other browsers.

Screen Shot 2020-10-14 at 10 03 13 AM

rlskoeser:
Ah, I see — it's the dashes themselves you're describing. That seems ok to me. If we care about this we could use an image, but I don't know how much it matters.

PR #146, opened 2020-10-19
Data weaving interface part 1 / footnotes refactor 

rlskoeser:
This is still in progress but I need some help & feedback on a few things.

You can see the current version here https://startwords-pr-146.onrender.com/issues/1/weaving-as-interface/ 

@thatbudakguy notes/questions/help needed:
- I need to disable scrolling when contextual notes are positioned and adjust where they are positioned vertically (either via css or js).
-  I think the notes js should have a proper js api so it can be called programmatically (right now I'm triggering it by locations, which seems kind of janky); but I'm not sure the best way to do that or how difficult it is. Feasible? Out of scope for now?
- preliminary dark mode styles (not yet used with dark-mode preferred property); would appreciate feedback on how to manage it better, seemed like I needed a lot of pretty specific selectors
- sometimes the endnotes aren't triggering, but I'm not sure why. thoughts? possible workaround: add a way for users to trigger them ('show endnotes'). Not ideal, obviously!
- long contextual notes aren't scrollable; I think it is specific to this page, presumably something with the deepzoom viewer. any ideas?

@gissoo :
- I would like to remove/disable the backlinks from the endnotes. I don't think there's an easy way to make them work.
- Should we link to the endnotes to the books on the Derrida's Margins site ? or no?
- If end note numbers should only be displayed if they can be in both views I vote we hide them. Are you ok with this?

Note that this update includes revised tag styles.


gissoo:
@rlskoeser 
re: 
- "I would like to remove/disable the backlinks from the endnotes. I don't think there's an easy way to make them work." – that's fine! 

- "Should we link to the endnotes to the books on the Derrida's Margins site ? or no?" – I think we should! I meant to write about this in the design issue and forgot. I think we should also link the contextual notes to the books on the site.  

- "If end note numbers should only be displayed if they can be in both views I vote we hide them. Are you ok with this?" – I'm not sure I'm following, but if I understand you correctly, its significance does not have anything to do with whether or not you can view both at once. I think they are important because it gives a clearer clue about which contextual note each end note is referring to. I think they should always be displayed in both places.

rlskoeser:
@gissoo thanks for responding.

Where in the notes should we link to the books on the Derrida's Margins site? Should the title be the link? 

Regarding the numbers: I communicated poorly! I'm not sure how to add the numbers to the zones on the zoom view, it might be hard to do. I'll see if I can add them somehow similar to the way they're generated for the end notes.

rlskoeser:
@gissoo I was able to add the numbers to the annotated weaving image! Not that hard after all. You can check them on the render PR site at your convenience.

If you want to revise the highlight zones to make them more consistent / line up better, you can edit them in recogito — I have a script that takes the recogito CSV download and turns it into the format I'm using here. Just let me know so I can download and regenerate after you've made changes.

gissoo:
> @gissoo thanks for responding.
> 
> Where in the notes should we link to the books on the Derrida's Margins site? Should the title be the link?
> 

- I was thinking of the title to be the link. 

> Regarding the numbers: I communicated poorly! I'm not sure how to add the numbers to the zones on the zoom view, it might be hard to do. I'll see if I can add them somehow similar to the way they're generated for the end notes.
> 

- Thank you!!

gissoo:
> @gissoo I was able to add the numbers to the annotated weaving image! Not that hard after all. You can check them on the render PR site at your convenience.
> 

- Yay, awesome!! Thanks!! 

> If you want to revise the highlight zones to make them more consistent / line up better, you can edit them in recogito — I have a script that takes the recogito CSV download and turns it into the format I'm using here. Just let me know so I can download and regenerate after you've made changes.
> 

- wow!! Ok, sure I'll let you know.

gissoo:
@rlskoeser I edited the highlight zones in recogito. Thank you!!

rlskoeser:
@gissoo thank you for editing the highlight zones! I've downloaded and regenerated my annotations file.

I also updated the notes so the titles like to the Derrida's Margins site, but I need you to decide how you want to style links in contextual notes and end notes.

gissoo:
@rlskoeser [I just revised them here](https://www.figma.com/file/MEt3gfKX4N3WTLPbLGBbGC/Startwords?node-id=1155%3A7350) 

Please let me know if you want me to test the annotations – they don't work when I go to the startwords site.

rlskoeser:
@gissoo added the link colors, thank you! You can see the updates here if you want to check the revised annotation zones: https://startwords-pr-146.onrender.com/issues/1/weaving-as-interface/

content/issues/1/data-beyond-vision/index.md
```diff
@@ -11,6 +11,7 @@ date: 2020-10-01
 doi: 10.5281/zenodo.3713671
 pdf: https://cdh.princeton.edu/media/uploads/documents/grad_fellows_call_for_spring_21.pdf
 images: ["issues/1/data-beyond-vision/images/dbv-social.jpg"]
+tags: [DataBeyondVision]
```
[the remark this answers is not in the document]
thatbudakguy: oh, just realized this is on the main DBV essay. I added this tag as part of #150.

content/issues/1/weaving-interface/index.md
```diff
@@ -0,0 +1,215 @@
...
+authors:
+  - Rebecca Sutton Koeser
+  - Gissoo Doroudian
+tags: [DataBeyondVision, Experiment]
+needs_deepzoom: true
+body_class: darkmode
+---
+
+
+This weaving represents the references in chapter one of Jacques Derrida’s de la Grammatologie (1967). The references have been cataloged and categorized by the research team of [Derrida’s Margins](https://derridas-margins.princeton.edu/). Each type of reference (epigraph, citation, quotation, footnote) is represented by a distinct yarn and weaving pattern. For more details on the goals and insights of this data physicalization, read Weaving Derrida’s References.
```
thatbudakguy, line 14: italicize _de la Grammatologie_?

thatbudakguy: can also link to the weaving snippet once #159 is merged

content/issues/1/weaving-interface/style.css
```diff
@@ -0,0 +1,198 @@
...
+    width: 100vw;
+    height: 230px;
+    margin-top: 15vh;
+    margin-bottom: 30vh;
+    background-color: black;
+
+    counter-reset: highlightZoneCounter;
+}
+
+/* breakpoint-s */
```
thatbudakguy, line 20: could always put these breakpoint values in the CSS vars block up top too

themes/startwords/assets/scss/_utilities.scss
```diff
@@ -3,7 +3,7 @@
 // gridded sections use a 10-column grid with 10px gaps
 .grid {
     display: grid;
-    grid-template-columns: repeat(10, 1fr);
+    grid-template-columns: [full-start] repeat(10, 1fr) [full-end];
```
thatbudakguy, line 6: whoa, I've never seen this syntax. what is it?

content/issues/1/weaving-interface/annotated-zoom.js
```diff
@@ -0,0 +1,72 @@
...
+
+        // create elements with page numbers to be positioned as overlays
+        var pageDiv = document.getElementById('page_numbers');
+        data.overlays.filter(el => el.className == 'page_n')
+            .forEach(el => {
+                const pageSpan = document.createElement('span');
+                pageSpan.innerText = el.content;
+                pageSpan.setAttribute('id', el.id);
+                pageDiv.appendChild(pageSpan);
+            });
```
thatbudakguy, line 26: likely doesn't matter when there are so few elements to be added, but i think it might be slightly more performant to batch everything by creating all the spans beforehand as a string and then setting `pageDiv.innerHTML` or similar.

content/issues/1/weaving-interface/annotated-zoom.js
```diff
@@ -0,0 +1,72 @@
...
+            id:"dataweaving-zoom",
+            prefixUrl:"https://cdn.jsdelivr.net/npm/openseadragon@2.4/build/openseadragon/images/",
+            preserveViewport: true,
+            visibilityRatio: 1,
+            // disallow zoom in/out
+            minZoomLevel: 4,
+            maxZoomLevel: 4,
+            defaultZoomLevel: 4,
+            // disable vertical panning
+            panVertical: false,
```
thatbudakguy, line 38: this combination of settings for me resulted in not seeing the numbers above (or footnotes below) the weaving on page load - I only discovered them because I used the red box in the preview overlay to pan around.

thatbudakguy (commented):
The dark mode CSS looks great on this, and it's pretty functional for me too.

> I need to disable scrolling when contextual notes are positioned and adjust where they are positioned vertically (either via css or js).

For CSS I usually do this with `pointer-events: none;` to prevent the element from being interactive (often applied to things that are overlays but shouldn't block things under them from being interactive). There might be settings besides `none` that are useful here, but I seem to recall it being a pretty restrictive property.

For JS, you could maybe do something like adding a scroll handler that captures and ignores scroll events when the note is open, and removing the handler when the note is closed. A nice bonus is that you could get more granular by picking the event to listen to - it would be awesome to allow horizontal scrolling within the OpenSeadragon viewer (including while notes are open) but disable vertical scrolling.

> I think the notes js should have a proper js api so it can be called programmatically (right now I'm triggering it by locations, which seems kind of janky); but I'm not sure the best way to do that or how difficult it is. Feasible? Out of scope for now?

Do you mean the contextual note JS more generally, or just the stuff for this piece? If the latter, I think you could benefit a little by moving stuff into a class, but it doesn't seem like an immediate/huge win - assuming this is what you mean by a proper API. If the former, definitely seems worthwhile - if only so we have it for usage on another project - but also not necessary anytime soon (unless you're already in debugging hell).

> preliminary dark mode styles (not yet used with dark-mode preferred property); would appreciate feedback on how to manage it better, seemed like I needed a lot of pretty specific selectors

Most of this is probably my fault from how I've written styles elsewhere, but I think it falls into two categories:

1. Just something dumb I did that can be simplified, e.g. `.footnotes ol li`. We'd probably never have a `.footnotes ul li` since footnotes are always numbered, so `.footnotes li` would've been fine and easier to override. Feel free to change this wherever you see it.
2. Something I did intentionally, e.g. `body.article`. The element (`body`) does add more specificity to the selector, but I like adding it in some cases so that the CSS actually reads more idiomatically - `.article` could be basically any element, and it's easier for me to understand quickly that `body.article` means "when the page body is an article". Selectors are longer (like `body.article.darkmode`), but in this case I think it's worthwhile (but interested if you agree).

> sometimes the endnotes aren't triggering, but I'm not sure why. thoughts? possible workaround: add a way for users to trigger them ('show endnotes'). Not ideal, obviously!

I didn't run into this issue, fwiw, but I feel like it's happened to me once or twice on the main DBV piece. I think your "show endnotes" solution is totally viable! You could just put the whole thing into a [disclosure widget](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary) - no JS necessary - and then there's no visual lag/layout shift when they appear as you scroll.

> long contextual notes aren't scrollable; I think it is specific to this page, presumably something with the deepzoom viewer. any ideas?

This I definitely did encounter. Not sure what's different about this page vs. DBV. Presuming you tried the `overscroll-behavior` thing we've used elsewhere. One thought I had (which is probably unhelpful at this point) is that if you keep the zooming disabled on OpenSeaDragon, there's somewhat less reason to use it (except for the nice little preview window). You could get most of the same functionality by just adding a flat image (that can be resized by hugo, even) that scrolls horizontally, and has all the notes connected to it via an imagemap. Then you can precisely control their locations, they are tabbable, etc. idk if this accomplishes everything you want, but I did find myself thinking "this just feels like an image map, but I can't scroll around in it..."

thatbudakguy:
forgot to mention - if it's helpful, I have a local version of this rebased against `main` with the conflicts already resolved which I used to test it out. so I can push that if you like

content/issues/1/weaving-interface/style.css
rlskoeser: I tried that and for some reason the variables didn't work in the media queries! I didn't mess with it for very long or investigate, though.

thatbudakguy: i kinda figured you had tried and it was just some syntax thing. let's not worry about it for now

themes/startwords/assets/scss/_utilities.scss
rlskoeser: I found it when I was trying to find a better way to do full-width in a grid! It lets you name a region and apply it. 

I created a new `full-width` class in the article scss that uses it like this:
```css
.full-width {
    width: 100vw;
    margin-left: (-$gutter);
    grid-column: full;
}
```

It made me wonder if we could be leveraging grid better / more semantically ...

content/issues/1/weaving-interface/annotated-zoom.js
rlskoeser: Good to know — this page is a little slow so it could help. I think I'd forgotten you could do that. Do you have an example/reference handy?

content/issues/1/weaving-interface/annotated-zoom.js
rlskoeser: Yes, I know it doesn't look right at some screen sizes  — my plan was to work on that with @gissoo and add some media queries to set the appropriate zoom level for the screen size.

Gissoo had proposed having a slider to control zooming in and out, which I was able to prototype, but there were some complications with getting it to work with the other controls and interactions. My inclination is to keep the interaction simpler if we can make it work.

rlskoeser:
> forgot to mention - if it's helpful, I have a local version of this rebased against `main` with the conflicts already resolved, which I used to test it out. so I can push that if you like

@thatbudakguy yes please!

content/issues/1/weaving-interface/annotated-zoom.js
thatbudakguy: it's something i unfortunately read about _after_ having built stuff like our s&co search, which takes the same approach you used above. lemme see if i can dig up the thing I read

thatbudakguy: so `.innerHTML` does work, but I think the approach I was actually thinking of was [this one detailed in MDN documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Performance_best_practices_for_Firefox_fe_engineers#Adding_nodes_using_DocumentFragments), which could be a pretty simple change to what you have. basically it just uses `DocumentFragment`, a node type that's designed to be more performant for use cases like this. their example maps fairly well here, i think.

rlskoeser:
@thatbudakguy thank you for the detailed review and responses!  

I was not clear enough when I asked about scrolling! I apologize. (I forget you aren't as familiar with the contextual notes code & logic as I am 🤦‍♀️ ).  The contextual note code has logic to scroll the page to make sure a note is fully visible — which I would like to turn off. I also want to override or customize the way it positions notes relative to the references. This is why I was asking about a js api for the contextual notes.

> Do you mean the contextual note JS more generally, or just the stuff for this piece? If the latter, I think you could benefit a little by moving stuff into a class, but it doesn't seem like an immediate/huge win - assuming this is what you mean by a proper API. If the former, definitely seems worthwhile - if only so we have it for usage on another project - but also not necessary anytime soon (unless you're already in debugging hell).

I meant the contextual notes code. (Sorry again for lack of clarity!). I would like to be able to call methods or configure it from the code for this piece — e.g. to turn off scrolling. There's already part of the code that turns it off internally, I just couldn't figure out how to do it from here (here = this piece).

If I move the contextual note logic into a class, what do I need to do to make that class be available for this code to interact with? Do you know of any good examples for this?

thatbudakguy:
> If I move the contextual note logic into a class, what do I need to do to make that class be available for this code to interact with? Do you know of any good examples for this?

What I would do is:
- move all code that should run once on page load into the init method for the new class
- move everything else into methods
- in a JS file that runs on every page, initialize the class
- bind the instance to a property on `window`, e.g. `window.notes`
- call the methods you need in the weaving interface js, e.g. `window.notes.position(foo)`

content/issues/1/weaving-interface/annotated-zoom.js
rlskoeser: Good to know about document fragment! And easy to adapt to use it here.

rlskoeser:
@thatbudakguy please review the contextual notes refactor. 

I'm not sure what's causing the endnotes not to display, sometimes when I'm testing and reloading locally the notes code isn'g getting initialized. (This is why I left the init logging in.) Could the `async` on the script tag have anything to do with this?

rlskoeser:
Forgot to say, in addition to refactoring I fixed the behavior for notes near the end of a document. RM's essay has one (I _knew_ it was going to happen somewhere). It isn't perfect, but I think it's better than it was.

themes/startwords/assets/js/notes.js
```diff
@@ -1,170 +1,243 @@
...
+                ContextualNotes.showEndnotes()
+                ContextualNotes.showEndnotesOnClose = false;
+            }
+        }
+    }
 
-window.addEventListener('hashchange', function() {
-    setupNote();
-}, false);
+    static noteCloseLinkHandler() {
```
thatbudakguy, line 118: think the JS convention is to name handlers things like `onNoteClose`, fwiw

themes/startwords/assets/js/notes.js
```diff
@@ -1,170 +1,243 @@
...
+        }
 
-/* on scroll, if a footnote is currently targeted, unselect */
-window.addEventListener('scroll', function() {
-    if (location.hash.startsWith('#fn:') && hideOnScroll) {
-        // close the note and unbind close link handler
-        closeNote(getCurrentNote());
+        if (location.hash == '#' + note.getAttribute('id')) {
+            event.preventDefault();
+            event.stopPropagation();
```
thatbudakguy, line 130: linter is giving me a warning I don't totally understand here, but I think it's to the effect of:
- this reference to `event` would be `undefined`...
- but it works, because of a (deprecated) spec that allows you to access the event anyway
- you can resolve it (i think) by just explicitly listing `event` as an argument to this handler

thatbudakguy: tried that and it seems to work.

thatbudakguy (approved):
refactor is looking good! only overarching comment is that, where you can, I think it's usually good to store and reuse references to DOM nodes instead of re-making `querySelector()` or other calls. so maybe the footnotes class could query for the whole block of footnotes (`<section>`) and the collection of actual `<li>`s and save them as instance variables in the constructor, and then your handlers can reference those instead of querying again.

lmk if anything doesn't make sense!

thatbudakguy:
re: `async` - I don't think so, but you can definitely switch it to `defer` and see what happens! if it doesn't have at least `defer` I think it'll fail, because it needs the whole DOM rendered to work.

themes/startwords/assets/js/notes.js
rlskoeser: Ah, good to know. This is the handler for the x link which is one way to close the note and calls the `closeNote` method. I feel like this name might be confusing — thoughts?

thatbudakguy: yeah, idk what name works best...`click` doesn't cover touch devices, and `link` doesn't totally explain that it's the "X" button (which really is more of a button here). might just need to add docblocks for each method that explain when they are called.

fwiw handler-named functions like `onNoteClose` do usually imply that they happen _prior_ to others, so I would expect something named `onNoteClose()` to be called before the actual call to `closeNote()` (which is indeed what happens).

rlskoeser:
@thatbudakguy I've renamed the method and added the missing event parameter. Are you comfortable merging it with those changes and I'll address your other comments with my remaining work on the annotated weaving? Or is there anything else you want to see fixed/handled before we merge this?

thatbudakguy:
nope, go for it - those were all just suggestions

Issue #149, opened 2020-10-19
fix pull-quote PDF rendering


thatbudakguy:
depends on #148

try using `page-break-inside: avoid`

rlskoeser:
fixed when the pagedjs update was merged, have confirmed fix on the site

Issue #148, opened 2020-10-19
change pdf generation workflow to use pagedjs


thatbudakguy:
as part of #88 we're exploring using [pagedjs](https://www.pagedjs.org/about/) instead of WeasyPrint since its output is more accurate using our print css.

rlskoeser:
closing this since the changes have been reviewed and merged in

PR #158, opened 2020-10-22
Chore/print styles redux

rlskoeser:
Revises print styles for PDF generation with pagedjs instead of weasyprint. 

todo:
- [x] fix endnote numbering in pdf
- [x] fix font size for pull quotes

ref #148 #149 

- Fixes the running header problem. Not sure if it is a bug or just a difference of implementation, but it only showed them on the last page because they were at the end of the document, so I moved them closer to the beginning. (I think it happened because running header content like this can change based on the content in the document.)
- Adjusts the pull quote styles: set to avoid page break inside, made wider to accommodate the long quotes in RM's essay, and adjusted the way the images were added in hopes of solving a weird problem (didn't help but I thought this solution was better).
- Added brief notes to the theme readme about generating PDFs
- Added css for bookmarks (PDF table of contents), but pagedjs doesn't support them. We might want to consider adding them manually for DBV (if feasible / not too difficult), since it's such a monster.

@thatbudakguy I turned off an `overflow:hidden` on article paragraphs because it was resulting in paragraph text not wrapping around pull quotes. Flagging for you because I'm not sure the implications.

@thatbudakguy I'm getting a timeout trying to generate the DBV essay. Have you run into this? Any ideas?

@gwijthoff I moved one of the pull quotes in RM's essay to make it work better in the PDF. I think you may have to futz with pull quote placement to get good PDFs. Is this acceptable? Here's a screenshot of what I'm seeing when a pull quote is near a page boundary. The avoid page break puts it on the next page, but there's a lot of whitespace and for some reason the corner images get repeated.
Screen Shot 2020-10-22 at 8 51 54 AM

@gissoo Would you check that the pull quote font size is correct? I think either it's too big or the quotes that Grant chose are longer than you anticipated. As an attempt to handle it better, I set pull quotes to 2/3 width instead of 1/2 — making them shorter vertically makes it much easier for the PDF generation to flow the content without big whitespace where it puts them on the next page. Here's a new PDF of RM's essay for you to check: [tdo.pdf](https://github.com/Princeton-CDH/startwords/files/5423064/tdo.pdf)


gwijthoff:
The pull quote placement looks great in PDF, @rlskoeser! Two lingering issues that I can see:

- endnotes are all numbered "0" in the PDF, but numbered correctly on the site
- more importantly, is there any way to avoid the large padding beneath the pullquotes on the site? When I narrow my browser window to mobile size, it creates a huge amount of whitespace beneath the pullquote that takes a while to scroll past. Here are screenshots of laptop and mobile. (It looks like these changes are already live on the site, given the whitespace beneath the pullquotes at our URL now? have portions of this PR been merged?)

Screen Shot 2020-10-22 at 1 28 33 PM

IMG_42D9D4DA1E56-1

thatbudakguy:
> I turned off an overflow:hidden on article paragraphs because it was resulting in paragraph text not wrapping around pull quotes. Flagging for you because I'm not sure the implications.

The implications of this are that underlines under contextual notes can sometimes push outside the paragraph boundary, which either makes them go into the margin or create horizontal scrollbars. Confirmed that this is happening in this branch:

Screen Shot 2020-10-22 at 1 44 48 PM

This might be an acceptable consequence though! It doesn't look too bad.

Regarding @gwijthoff's post above, I didn't see any of these whitespace issues for any device width (tested on FF on my macbook using the mobile device emulator). Everything looks great.

themes/startwords/assets/scss/print.scss
```diff
@@ -226,6 +224,22 @@ sup {
...
+    page-break-inside: avoid;
+    bookmark-label: content(text);
+}
+h1 { bookmark-level: 1; }
+h2 { bookmark-level: 2; }
+h3 { bookmark-level: 3; }
+h4 { bookmark-level: 4; }
+h5 { bookmark-level: 5; }
+h6 { bookmark-level: 6; }
+
```
thatbudakguy, line 242: Wow, this is cool to know about!!

themes/startwords/layouts/article/single.html
```diff
@@ -55,24 +55,25 @@ <h1>{{ .Params.Title }}</h1>
...
+                <a class="first-page-header" href="{{ relURL "/" }}" aria-label="Startwords">
+                    <img alt="Startwords" src="{{ relURL "/pdf-logotype.svg" }}" /></a>
+
+                <a class="page-header" href="{{ relURL "/" }}" aria-label="Startwords">
+                    <img alt="Startwords" src="{{ relURL "/pdf-logo.svg" }}" /></a>
+
+                 <a href="http://doi.org/{{ .Params.doi }}" rel="alternate" class="page-doi">doi:{{ .Params.doi }}</a>
+
+            </section>
+
```
thatbudakguy, line 71: worth moving this into its own template partial? or is there no point

thatbudakguy (approved):

gwijthoff:
Wow, no one else is seeing that huge whitespace beneath pullquotes on the live site right now? I've tested it on my macbook in chrome/ff/safari and on my phone in ff/safari and get the same results. @thatbudakguy @rlskoeser @gissoo

thatbudakguy:
@gwijthoff I see it on the live site, but not in this branch - sorry for not specifying

gwijthoff:
@thatbudakguy no problem. The extra whitespace is also appearing when I serve this branch locally, in all browsers, which is how I first noticed it.

themes/startwords/assets/scss/print.scss
rlskoeser: I know! Wish it was supported by paged.js, but maybe they will add it at some point

themes/startwords/layouts/article/single.html
rlskoeser: I don't think so? Not used anywhere else right now

rlskoeser:
@thatbudakguy thanks for explaining the implications of the overflow change — I don't think it looks too bad to have the contextual note underlines expanding, and it's pretty important that the text wrap around pull quotes!

@gwijthoff I missed the problem with footnote numbering, thanks for flagging! Will have to check what's going on there. And I'll definitely make sure to pull your edits and resolve changes before merging (thanks for letting me know).

rlskoeser:
@thatbudakguy any ideas about the 0 counter on the footnotes? pagedjs documentation includes an example exactly like what we're doing this so I'm confused about why it doesn't work properly https://www.pagedjs.org/documentation/06-generated-content/

thatbudakguy:
> any ideas about the 0 counter on the footnotes? pagedjs documentation includes an example exactly like what we're doing this so I'm confused about why it doesn't work properly https://www.pagedjs.org/documentation/06-generated-content/

been banging my head against this for awhile, not really much progress. I did though learn there's a [whole footnote spec](https://www.w3.org/TR/css-gcpm-3/#footnotes) in the works...which evidently [neither weasyprint nor pagedjs](https://medium.com/printcss/printcss-footnotes-9bb67fb2064b) support as of october 13. fun times!

rlskoeser:
@thatbudakguy if we can't get the counters to work maybe I can figure out a way to use the default `ol` markers for print only.

rlskoeser:
@gissoo Could you weigh in on the pull quotes regarding font size and width? You may have missed my question earlier, it was buried under a lot of other questions and comments! Here's what I wrote:
> Would you check that the pull quote font size is correct? I think either it's too big or the quotes that Grant chose are longer than you anticipated. As an attempt to handle it better, I set pull quotes to 2/3 width instead of 1/2 — making them shorter vertically makes it much easier for the PDF generation to flow the content without big whitespace where it puts them on the next page. 

You can look at them in this test pdf: [tdo.pdf](https://github.com/Princeton-CDH/startwords/files/5423064/tdo.pdf)

rlskoeser:
I can't get the counters to work in pagedjs, even though it's in their documentation and their example is exactly what we're doing.  I found some reported bugs related to this functionality, but it seems like they've been fixed.

If we can figure out where, we should ask a question to figure out what's wrong in our case, but in the short term — I've revised print styles to use the list markers.

I can't style them to match the designs exactly, I hope this is close enough (screenshot / pdf):
Screen Shot 2020-10-23 at 11 05 38 AM

[tdo.pdf](https://github.com/Princeton-CDH/startwords/files/5430260/tdo.pdf)

gissoo:
> @gissoo Could you weigh in on the pull quotes regarding font size and width? You may have missed my question earlier, it was buried under a lot of other questions and comments! Here's what I wrote:
> 
> > Would you check that the pull quote font size is correct? I think either it's too big or the quotes that Grant chose are longer than you anticipated. As an attempt to handle it better, I set pull quotes to 2/3 width instead of 1/2 — making them shorter vertically makes it much easier for the PDF generation to flow the content without big whitespace where it puts them on the next page.
> 
> You can look at them in this test pdf: [tdo.pdf](https://github.com/Princeton-CDH/startwords/files/5423064/tdo.pdf)
>

- @rlskoeser sorry I missed your question earlier. I just checked both the site and the pdf 
1. the pull quote font sizes on the site are correct, except I don't see any "line-height" property, is there another name for it in CSS that I can't find? It should be 30 on mobile and tablet, and 32 on desktop. I'm not sure if it is right now, looks a little tight, if it is then never mind. 

2. the pull quote font sizes on the pdf are not proportional to the main body's font sizes, which is why it looks large right now. Do you remember what the other article's main body font size was for the pdf? We would want to match that one – the body in this PDF is 9p (12px) and if this the correct body size and if we keep it this way then the pull quote would need to be 12p (16px) – **the pull quote is always 4px larger than the main body.** 

3. Why is there an empty spot below the pull quotes? It doesn't match the designs, which wrapped the text around the pull quote. It also looks really awkward right now :( 

Please let me know if you have any questions.

rlskoeser:
Thank you @gissoo ! Your feedback was so clear and helpful once I sat down to make this change.

- I added the line heights you specified. You can check here: https://startwords-pr-158.onrender.com/issues/1/their-data-ourselves/
- Adjusted the font size for print/pdf as you said (it is indeed 12px for body and so 16px for quotes). They look much better and are flowing better within the pages of the document.  Here's a revised pdf to check: [tdo.pdf](https://github.com/Princeton-CDH/startwords/files/5431143/tdo.pdf)

Does the line height for pull quotes need to be adjusted any for print?

gissoo:
> Thank you @gissoo ! Your feedback was so clear and helpful once I sat down to make this change.
> 
>     * I added the line heights you specified. You can check here: https://startwords-pr-158.onrender.com/issues/1/their-data-ourselves/
> 
>     * Adjusted the font size for print/pdf as you said (it is indeed 12px for body and so 16px for quotes). They look much better and are flowing better within the pages of the document.  Here's a revised pdf to check: [tdo.pdf](https://github.com/Princeton-CDH/startwords/files/5431143/tdo.pdf)
> 
> 
> Does the line height for pull quotes need to be adjusted any for print?
>

- @rlskoeser thank you!! These look great!! So much better!! Yes, on the PDF, the line height is very high right now, as you said it's 30px, the pull quotes have a 16px font size so a line height of 21px should look appropriate for this based on the decisions made also for the web version, if you find 21px too tight then 22px would be ok too.

rlskoeser:
Thank you @gissoo ! 21 looked tight to me so I went with 22. Here's a screenshot of how it looks now:

Screen Shot 2020-10-23 at 3 04 23 PM


Are we good to merge now?

gissoo:
@rlskoeser perfect!! Thank you!! Yes please!!

Issue #147, opened 2020-10-19
pdf styles for DBV essay


thatbudakguy:
broken off from #88.

PR #160, opened 2020-10-22
PDF styles for DBV essay (#147)

thatbudakguy:
Current status of the pagedjs workflow still isn't where we want it to be, but you _can_ get a workable PDF - see below.

# changes
This PR contains a mishmash of changes that I made while I was trying to figure out why pagedJS hangs rendering the PDF; I initially thought it might be due to malformed HTML. Validating the HTML led me to a lot of errors, which I resolved, but that didn't end up fixing the rendering issue. These include:

- `<script>` tags don't need (and shouldn't have) a `type` attribute if they're plain javascript, so I removed them.
- the `figure` shortcode had a bunch of formatting problems and wasn't actually parsing our `srcset` because percentage values aren't allowed in it. I fixed those issues.
- the `wrap` shortcode could render an empty `id=` attribute, causing many (40+) elements to have the same (empty) `id`. I edited it to use `with` so this doesn't happen.
- if you write two shortcodes in a row without creating empty space between them, hugo creates a single `</p>` tag without an accompanying opening tag. I _think_ browsers ignore this, but I added spacing just in case.

# the pdf
I still can't find what exactly triggers pagedjs to hang; I've tried rendering the essay without the sketchfab/deep zoom embeds and that didn't seem to help. The way I rendered the PDF in this PR was to run:
```
$ hugo server -D
$ npx pagedjs-cli http://localhost:1313/issues/1/data-beyond-vision -d
```
The `-d` flag opens the chromium instance that pagedjs uses via puppeteer to do rendering and lets you interact with it. Once the window is open, it will show you the essay but will appear to not have print styles applied. If you scroll down "far enough" - right around the weaving portion of the essay for me - the print styles will suddenly be applied to everything. Continue scrolling until you reach the bottom, and the pagedjs CLI will report "26 pages rendered". Now, you can actually generate the PDF by using Chrome's print dialog (in the debug window), choosing "print to pdf", and making sure "render background graphics" is checked.

It's not a great workflow, but it does at least allow you to preview the rendered PDF! 

My thinking is that pagedJS does the rendering page-by-page (digging into the source supports this), so _something_ on one of the later pages is calling it to hang indefinitely. Forcing the browser to load the page allows pagedjs to "skip" past its issue and lets the whole thing get rendered.


thatbudakguy:
other note: I couldn't get the icon navigation sections ("choose a section") to render properly, so I ended up just hiding them for the PDF version. I couldn't find any errors in the console that would suggest why they rendered incorrectly, and it seemed to be inconsistent - they showed up in some places and just didn't appear at all in others. adding `page-break-inside: avoid;` didn't seem to help.

it's a bummer not to have them in the PDF, but I think they're a little bit more useful in the web version as a kind of navigation tool, so I think we might be able to live without them for the PDF.

rlskoeser:
@thatbudakguy wow, thanks for all the work figuring out a way to generate a pdf and work around the pagedjs timeout issue. It's annoying we have to do it that way, but hopefully we can open an issue and get some feedback about what's causing the problem.

The css changes you've made look good to me.  

I commented out the line that hides the `.icon-nav` because I wanted to see what was going on, and it rendered just fine for me. Would prefer to keep them if we can! Chromium version 86.0.4240.0 if it matters.

A few things I noticed:
- The pie chart is enormous again, can we make it smaller for print? Maybe the bar chart too (although that one bothers me less than the gigantic pie chart)
- The top of the weaving icon next to the title "weaving Derrida's References" is cut off
- The font size for the tags at the top of the article looks big. IDK if we have tag font size for PDF, but I think it's the same as other header font sizes.
- I wonder if we should have a directive to avoid page breaks inside footnotes? One of them has an image that wraps to the next page. It's not terrible, but might be a little better keeping them together.

thatbudakguy:
> I commented out the line that hides the .icon-nav because I wanted to see what was going on, and it rendered just fine for me.

Tried again and...now it does for me too? No idea what happened, but happy we get to keep them!

> The pie chart is enormous again, can we make it smaller for print?

This was a combo of a regression in the `figure` shortcode (nice catch) and the fact that it also just looks bigger in print. I fixed the regression and also just made it a little smaller for good measure, since it doesn't ever need to be all that big.

> The top of the weaving icon next to the title "weaving Derrida's References" is cut off

Added margin to address this

> The font size for the tags at the top of the article looks big.

Yep, just needed to add a selector - should be accurately sized now

> I wonder if we should have a directive to avoid page breaks inside footnotes?

Tried this and it seems to work well, so I kept it.

rlskoeser:
@thatbudakguy do you want further review? Happy to approve this so you can merge it in

rlskoeser (approved):

PR #161, opened 2020-10-22
revert baseURL to full form

thatbudakguy:
testing something out:

- revert the hugo config to use our full startwords.cdh.princeton.edu URL as the `baseURL`
- tell render to set an environment variable (`HUGO_BASEURL`) that overrides this for builds and sets it to `/`


thatbudakguy:
success! the relative links work for the render preview above, and CI is passing.

rlskoeser (approved):
Looks great! Thanks for figuring out how to solve this

Issue #154, opened 2020-10-21
visiting anchor links cuts off top of page


thatbudakguy:
this appears to take effect any time you use an anchor link, including clicking the icons in the DBV section headers as well as visiting contextual notes/endnotes.

> The one endnote issue I've encountered comes when clicking the return arrow in the endnote to glide back to its anchored place in the text. When I then scroll to the top of the page, this is now the topmost place on the page (pictured below). I can no longer scroll any higher. Maybe not super important, but strange.
> <img width="796" alt="Screen Shot 2020-10-20 at 4 17 34 PM" src="https://user-images.githubusercontent.com/5580840/96639515-cd8d3800-12ef-11eb-87de-da618b60222d.png">

_Originally posted by @gwijthoff in https://github.com/Princeton-CDH/startwords/pull/152#issuecomment-713114575_

PR #162, opened 2020-10-22
Fix clipped spacing at article top (#154)

thatbudakguy:
This had a weird solution that I don't totally understand, but it does appear to work! Removing the spacing that pushed down the background scarf allowed the page to scroll up to the top. So, the issue was localized to the DBV essay only.

@gwijthoff check the render preview when it's finished and let me know if the problem is resolved for you too.


gwijthoff (approved):
Looks like that fixed it!

Also, bless you Render Bot.

Issue #145, opened 2020-10-14
adjust main content container to max line length of ~75 characters


thatbudakguy:
relevant styles:
https://github.com/Princeton-CDH/startwords/blob/00125b2243316c16837898a9eed5d048a9b27205/themes/startwords/assets/scss/_utilities.scss#L3-L35

PR #165, opened 2020-10-23
Prevent long lines in article body text (#145)

thatbudakguy:
- Creates a constrain-text mixin for limiting elements to <75char
- Applies the mixin to most textual elements in article bodies
- Applies the mixin to article headers and footnotes

See #145

### DIFF
--- a/themes/startwords/assets/scss/_utilities.scss
+++ b/themes/startwords/assets/scss/_utilities.scss
@@ -34,6 +34,14 @@
     }
 }
 
+// some text elements shouldn't be more than ~75 characters wide
+@mixin constrain-text {
+    margin-left: auto;
+    margin-right: auto;
+    width: 100%;
+    max-width: 32rem;
+}
+
 // inverted sections use a purple background & white text
 @mixin inverted {
     color: $off-white;
--- a/themes/startwords/assets/scss/_utilities.scss
+++ b/themes/startwords/assets/scss/_utilities.scss
@@ -34,12 +34,14 @@
     }
 }
 
-// some text elements shouldn't be more than ~75 characters wide
+// some text elements shouldn't be more than ~75 characters wide on desktop
 @mixin constrain-text {
-    margin-left: auto;
-    margin-right: auto;
-    width: 100%;
-    max-width: 32rem;
+    @media (min-width: $breakpoint-m) {
+        margin-left: auto;
+        margin-right: auto;
+        width: 100%;
+        max-width: 32rem;
+    }
 }
 
 // inverted sections use a purple background & white text
--- a/themes/startwords/assets/scss/article/_notes.scss
+++ b/themes/startwords/assets/scss/article/_notes.scss
@@ -1,6 +1,8 @@
 body.article .footnotes {
-    margin: rem(120px) 0;
     min-height: 50vh;
+    margin-top: rem(120px);
+    margin-bottom: rem(120px);
+    @include constrain-text;
 
     > hr { display: none; }
 
--- a/themes/startwords/assets/scss/article/_single.scss
+++ b/themes/startwords/assets/scss/article/_single.scss
@@ -1,8 +1,7 @@
 /* single article page styles */
 
 body.article {
-
-
+    
     article {
         margin-top: rem(60px);
         margin-bottom: rem(100px);
@@ -16,9 +15,12 @@ body.article {
             margin-bottom: rem(30px);
             @media (min-width: $breakpoint-s) { margin-bottom: rem(80px); }
             @media (min-width: $breakpoint-m) { margin-bottom: rem(60px); }
-
-            p { margin: 0; }
-
+            @include constrain-text;
+            
+            p {
+                margin: 0;
+            }
+            
             .number {
                 a, a:visited {
                     color: $dark-grey;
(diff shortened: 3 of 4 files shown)


gwijthoff:
Wow, those line lengths are gorgeous! Nice work.

My one concern is what this change does to pull quotes. They look a little off-kilter now, and wedging them between paragraphs (rather than wrapping the main text around the pull quote) makes the pullquote feel somehow equivalent to a flourished section break. Thoughts @gissoo?

Screen Shot 2020-10-23 at 1 55 21 PM

thatbudakguy:
I believe this is related to/the same as #149, and thus will be fixed when we merge #158.

thatbudakguy:
this is ready for another look now that pull-quotes have been fixed.

@gwijthoff I see what you mean about them looking off-kilter, but I actually kinda like it if you're OK with it - it helps stop the pullquotes from getting even taller, which was iirc was kind of a problem for PDF because it's harder to avoid page breaks that way. now that the text container is tighter, if we restrict them more they'll probably get pretty tall.

gwijthoff:
@thatbudakguy this looks perfect! Looking now, I don't have a problem with the pullquotes going wider than the main text line length. I actually prefer it. The main problem was the pullquote looking like a section break with the text didn't wrap around it, but that's resolved now, so we're all good

gissoo:
@thatbudakguy @gwijthoff to be honest I don’t like them not aligning with the paragraph, they look really odd that way :(

gwijthoff:
This isn't a completely accurate rendering, just playing with window width to see how different things would look, but one consideration is that allowing the pullquote to extend beyond the text width gives both quote and main text lines more width. Aligning the pullquote with main text width kind of smooshes both so that the pullquote text is sometimes only one word per line. 

Although I'm looking back at #4 and both of the pullquote examples I screenshotted from other publications align with the main text margins…

Screen Shot 2020-10-23 at 3 46 17 PM
Screen Shot 2020-10-23 at 3 51 20 PM

gissoo:
@gwijthoff thanks for writing. This is something we should have talked about before as the design was being reviewed, I understand how it's helping, however the problem is that extending it beyond the body right now at its current width is creating an imbalance on the page. If I had designed for it I would have made sure its width would be appropriate for this purpose.

@thatbudakguy would it only be extending beyond the paragraph width on desktop? 
What would it look like on tablet and mobile? 
And you know how there are currently 3 pull quotes on the article like this: left, right, left – if we extend them beyond the paragraph width do you know if they will still be off to the left, right, left?

gwijthoff:
@gissoo to answer the @thatbudakguy question, yes. That's how they currently appear in [the render preview](https://startwords-pr-165.onrender.com/issues/1/their-data-ourselves/), left-right-left beyond the margin.

gissoo:
@gwijthoff thanks for the link: It looks very wrong in these two screenshots: If we do choose to extend it beyond the paragraph then it would look best if 
1. it only happens on desktop and not on other views, 
2. only on desktop if it would adjust to avoid the blank space when the screen is wider on desktop. 
Screen Shot 2020-10-23 at 4 05 30 PM
Screen Shot 2020-10-23 at 4 06 21 PM

thatbudakguy:
> it only happens on desktop and not on other views

OK, I've restricted it to desktop. This does mean that some large tablet/phablet-like devices might have lines of text longer than 75 characters, but I don't think there are really a ton of devices in this category (would be widths between ~768px and 1024px, I think)

@gissoo I actually can't replicate either of the screenshots you posted...can you post browser info?

gissoo:
> > it only happens on desktop and not on other views
> 
> OK, I've restricted it to desktop. This does mean that some large tablet/phablet-like devices might have lines of text longer than 75 characters, but I don't think there are really a ton of devices in this category (would be widths between ~768px and 1024px, I think)
> 
> @gissoo I actually can't replicate either of the screenshots you posted...can you post browser info?
>

It's firefox desktop.

gissoo:
@thatbudakguy @gwijthoff I would vote to hold off on this and go with what we had already agreed on, I anticipate some alignment issues will come up with this as we will keep viewing it in various conditions which could be more problematic than what this might be solving ...just fyi

thatbudakguy:
Screen Shot 2020-10-23 at 16 47 17

@gissoo here's what it looks like for me at 1080p on firefox desktop. you're using the render preview link above, right?

gissoo:
@thatbudakguy yay! it looks SO much better …was making me seriously depressed – @budak the one on the left happens when you have the page already open and then expand it …but if you expand the page and refresh it is fixed …this is questionable since most people won’t refresh their pages when they expand them – can't anything be done about this? 
Screen Shot 2020-10-23 at 4 38 16 PM
Screen Shot 2020-10-23 at 4 43 50 PM

thatbudakguy:
> can't anything be done about this?

I doubt it, I think that's just the behavior of the rendering engine. Repaints/reflows are very expensive, so it would make sense for the browser to not constantly try to do it when you drag the window.

fwiw I don't think resizing the window is a good way to test - you should use the mobile device emulator or a real device whenever you can. trying to design for and code for resized windows introduces quite a lot of complexity, and we've got enough to handle just working across different devices. i don't even think this behavior is very common - people certainly resize windows on occasion, but nobody expects a site to look perfect if you squish it down really far or stretch it out (at least not without giving it a chance to refresh).

gissoo:
> > can't anything be done about this?
> 
> I doubt it, I think that's just the behavior of the rendering engine. Repaints/reflows are very expensive, so it would make sense for the browser to not constantly try to do it when you drag the window.
> 
> fwiw I don't think resizing the window is a good way to test - you should use the mobile device emulator or a real device whenever you can. trying to design for and code for resized windows introduces quite a lot of complexity, and we've got enough to handle just working across different devices. i don't even think this behavior is very common - people certainly resize windows on occasion, but nobody expects a site to look perfect if you squish it down really far or stretch it out (at least not without giving it a chance to refresh).
>

@thatbudakguy unfortunately it makes sense that browsers would do this. I do use the emulator, however fwiw the actual experience of the users involves the resizing of the windows a lot on desktop, and many don't know about refreshing it after resizing it!

thatbudakguy:
are we OK to merge this?

gwijthoff:
@thatbudakguy I was just reading back through the whole chain and it looks like it, yes

gissoo:
Please merge! Thanks for writing!!

PR #166, opened 2020-10-23
Add mediatype and crossorigin to font preload links

rlskoeser:


thatbudakguy:
looks good! I never tried adding `type`, maybe that was the secret sauce. we could probably preload even more on the article page, but this seems fine for now.

PR #167, opened 2020-10-26
Shorten snippet slugs

rlskoeser:
- configures short, consistent slugs so howto snippets have better urls
- updates links in DBV essay to the new urls; also made them relative to current site
- configured snippet order to match model order in DBV essay, experiment last
- removed versions from svg filenames and updated in DBV essay


rlskoeser:
Review from either @thatbudakguy or @gwijthoff is probably fine — although @gwijthoff I did wonder if you'd thought about snippet order.

content/issues/1/data-beyond-vision/index.md
```diff
@@ -209,7 +209,7 @@ The bar chart conveys a sense of certitude and exactness that does not reflect t
...
 
-[Make one >](https://startwords.cdh.princeton.edu/issues/1/print-a-model-of-the-shakespeare-and-company-lending-library-membership/)
+[Make one >](/issues/1/modeling-howto/)
```
thatbudakguy, line 212: are we for sure good with relative links for these? I used absolute URLs because I wanted the regex that expands them in the text-only output to create a link that would actually work out of context, but that might not be that important.

thatbudakguy (approved):
lgtm; it's good to have some consistency with the slug naming too. defer to @gwijthoff on order of pieces.

content/issues/1/data-beyond-vision/index.md
rlskoeser: I wondered why you were using absolute links! It def. makes it harder to test the links on a different instance.

Relative links in the text versions sounds like an editorial decision to me, defer to @gwijthoff . If we care about having full links, I'd much rather see a regex solution that adds the base url vs hardcoding it to the production site.

PR #168, opened 2020-10-26
Weaving interface pt2

rlskoeser:


thatbudakguy (commented):
still can't scroll inside contextual notes, but presumably you're working on a solution to that. 

it's a little annoying to not be able to scroll past the viewer because it swallows scroll events; i think john jameson mentioned this as a problem on one other review he did for us where we had a full-width viewer. not a dealbreaker, just kinda jarring when the page stops scrolling abruptly.

I'm confused about the usage of `static` for some but not all methods in the `ContextualNotes` class. If it's intended to operate like a singleton, where there's always one `ContextualNotes` and you use its API by talking to the class (which makes sense; can't imagine we'd ever want more than one instance) I'd just make all the methods static. Otherwise I'm not sure what it's doing, since making things like `closeNote()` and `openNote()` static means they don't get access to properties like `this.currentNote`.

One issue that became apparent is that in some situations (namely on pageload), `closeNote()` is called when:
- there's no note active
- there's nothing in the hash at all, because the page has just loaded

This is because (I think) the OpenSeadragon viewer fires off a few `pan` events...evidently just for fun. I see ~3 of them on pageload. The result is that `closeNote()` tries to `getElementById` on `undefined`.

To mitigate it, I would use a simpler idiom at the start of `closeNote()` and `openNote()`, like:
```js
note = note || this.currentNote
```

The way you've written `this.currentNote()` only tries to fetch an element if there's something in the hash, so I would just delegate to that wherever you can.

rlskoeser:
@thatbudakguy thanks for the feedback! I'd wondered about making all the methods static. it seems like it simplifies a lot of things.  Thanks for the catches and the suggestions.

The panning you're seeing OpenSeadragon do at the beginning is in my code — I'm setting it to display the left side of the image (it defaults to center), fitting the image vertically, and then zooming slightly based on that. If you have thoughts on doing that more efficiently, let me know! I experimented a little with defining a rectangle for coords I want shown, but taking advantage of `fitVertically` seemed like the most elegant and reliable solution.

thatbudakguy:
> The panning you're seeing OpenSeadragon do at the beginning is in my code

Oh, that totally makes sense. I guess a super quick fix then would be to temporarily disable `closeNoteOnScroll` or whatever that setting is called while you initially pan the viewer, then turn it back on.

rlskoeser:
@thatbudakguy have made the changes you recommended and addressed the other problems I was working on — do you want to review again or ok to merge?

PR #172, opened 2021-03-02
Use site dark purple for browser theme / background color

rlskoeser:
I've noticed this since I'm using vivaldi some — we had the red from the css set as the browser background / theme color and it doesn't look great. This switches it to dark purple.


thatbudakguy (approved):
lgtm; i don't think i ever used a browser that actually displayed this color - glad to have it tested!

Issue #176, opened 2021-05-28
single author / DOI running together on home page and issue list


rlskoeser:
**Describe the bug**
On desktop widths, for a feature essay with only one author, the author name runs into the DOI. Happens on both the [home page](https://startwords.cdh.princeton.edu/) and on the [issue page](https://startwords.cdh.princeton.edu/issues/1/)

**Screenshots**
![Screen Shot 2021-05-28 at 9 41 21 AM](https://user-images.githubusercontent.com/691231/119993909-1568f180-bf9a-11eb-9ae5-941b26d3280e.png)

rlskoeser:
fixed in PR #180

PR #180, opened 2021-07-14
Correct display for author list in summary view

rlskoeser:
fix for the single-author display problem documented in #176   

I guess this was just a typo in the style selector?


thatbudakguy (approved):
nice catch!

PR #194, opened 2021-12-01
Revise issue2 styles

rlskoeser:
updates include:
- center twitter cards on the Douglass Day essay
- adjust layout for iframes in articles (for the keyboards)
- revise keyboard styles to harmonize with SW colors & fonts, make clearer they are interactive
- kaltura video embed styles for serendipity piece


gissoo:
@rlskoeser thank you! These look much better now. Is there a way to make all the tan/light containers for the keyboard purple and instead add a line for the search section with a help text "type ..." like how we have it for S&Co (but a solid line instead of a dotted one
Screen Shot 2021-12-01 at 7 02 18 PM
) 

– also, I think after you changed the styles you have to scroll within the frame to see the output section, it's not showing the entire component in one view
Screen Shot 2021-12-01 at 7 04 19 PM
.

rlskoeser:
Thanks @gissoo . I added a placeholder text on the first form like you suggested, and removed the tan background. (I'm not sure the placeholder text makes as much sense in the later keyboards and hoped adding it to the first one would be enough.)

I knew about the vertical space issue — it should be fixed now for the online version. I thought I'd corrected it for the print version also, but there is still too much space in the PDFs.

PR #198, opened 2021-12-02
issue 2 pdf improvements

rlskoeser:
- add and document new `video` shortcode that should be used to wrap video content to provide preview image for print / pdf version 
- use video shortcode in community & serendipity essays
- add still image for serendipity video preview
- adjust twitter rendering in pdf (still not great)
- adjust pull quotes locations in data's destination essay

(should the shortcode have a different name?)


gwijthoff:
@rlskoeser it looks like the H2 headings are creating page breaks in the most recent PDF you uploaded to Drive. Both in snippets and features. Is that something we can resolve in this PR or should I create another?

rlskoeser:
@gwijthoff page break before H2 is what Gissoo designed and I implemented for issue 1. I've been wondering about it, though — it doesn't seem like it works well for the articles in issue 2, and it sounds like you agree with that. Should I make that a custom style for Data Beyond Vision and turn it off for articles generally? 

It should be a small change, seems fine to include in this PR (we can rename it).

rlskoeser:
@gwijthoff what about the short width for h2 titles? Should that be a DBV specific style? The wrapping looks strange for some of the headings in these essays.

gwijthoff:
@rlskoeser agreed that the page break before H2 should be a DBV-specific style. I also agree that the short-width line wrapping for H2s should be removed. The below looks odd. No reason to wrap that line any narrower than a regular line of body text.
Screen Shot 2021-12-03 at 10 55 07 AM

content/issues/2/datas-destinations/index.md
```diff
@@ -109,9 +109,8 @@ Boermans also supplied dated manuscript evidence, which means that “Taffety”
...
...y compared transcriptions produced by Zooniverse staff, who included an early modernist with paleography expertise (Van Hyning) as well as folks without paleography training, including astronomers and developers, many of whom hold higher degrees in other fields including philosophy, ecology, and computer science. {{<pullquote right `Thousands of volunteers transcribed, discussed, and conducted research about early modern manuscripts, history, language, and culture … [in] a form of collective close reading` >}} The project team also gathered transcription data from the pool of Zooniverse beta tester volunteers, a subset of the wider Zooniverse user community who test projects under development. Early aggregation results were promising, showing approximately 98 percent agreement between transcribers in many cases. But as the project went public and the heterogeneity of both the participants and the documents increased, clustering accuracy dipped, which negatively impacted text string comparison. In retrospect, it’s clear that our beta testing sample was too skewed towards people with higher degrees, relevant subject knowledge, and/or existing familiarity with the Zooniverse platform.
```
gwijthoff, line 113: @rlskoeser now that the pull quote isn't on its own line, it seems to affect the formatting of the body text

Screen Shot 2021-12-03 at 1 34 43 PM

rlskoeser:
@gwijthoff drat, I was only looking at the pdf. Ok to move between the next paragraphs?

gwijthoff:
@rlskoeser between the paragraphs ending and beginning "with the Zooniverse platform." and "From 2016 to 2018"? Sure! If so, in order to space them out more, can you then move the next pull quote containing "Each word on each page" between the paragraphs ending and beginning "those in EMMO." and "Both the successes and the challenges"?

rlskoeser:
btw @gwijthoff one of the pull quotes has part of a parenthetical (it has a close paren but no opening paren); was that intentional?

gwijthoff:
> btw @gwijthoff one of the pull quotes has part of a parenthetical (it has a close paren but no opening paren); was that intentional?

@rlskoeser good catch! Just removed that `)`

Issue #195, opened 2021-12-01
Issue 2 misc to-do's


gwijthoff:
- [x] #196
- [x] in PDF: page break after header / article metadata
- [x] in PDF: footnote numbers are duplicated, 11, 22, 33, etc
- [x] in PDF: keyboards, tweets, and videos have some display issues
- [x] in PDF: H2 headings create page breaks in all PDFs
- [x] in TXT: header ASCII art scrambled -- we think this will be resolved when publishing on GH Pages
- [x] in TXT: need a line break after metadata block and article text
- [x] Keyboards
    - [x] in PDF: add descriptive text about interactive elements: "The online version of this essay includes interactive keyboards threaded throughout the text."
    - [x] fix typos and format notes in keyboard interludes
    - [x] in PDF: fix extra vertical space for embedded keyboards
- [x] Serendipity
    - [x] TXT output includes both alt text and the embed code for the video
    - [x] TXT encoding doesn't work for Hebrew characters
    - [x] PDF replace video with alt still image and note about video
- [x] Douglass
    - [x] no TXT output for Tweets
    - [x] no TXT output for videos
    - [x] PDF replace youtube videos with alt still image and note about video (custom youtube shortcode?)
- [x] #197
- [x] generate PDFs for all four pieces for review by authors
- [x] one of the pull quotes on Data's Destinations is doing that funky page boundary thing we dealt with on RMs essay

kmcelwee:
@rlskoeser @gwijthoff I believe this can be closed

PR #200, opened 2021-12-06
Wrap notes in .sidenote class, fix line-height property in PDF, fix typos

kmcelwee:
Addresses some things in #195

### DIFF
--- a/themes/startwords/assets/scss/article/_figure.scss
+++ b/themes/startwords/assets/scss/article/_figure.scss
@@ -9,7 +9,6 @@ body.article .text-container > p > img {
 
 body.article figure {
     margin: rem(30px) 0;
-    page-break-inside: avoid;
 
     img {
         display: block;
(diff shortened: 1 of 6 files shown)


themes/startwords/assets/scss/article/_figure.scss
```diff
@@ -9,6 +9,7 @@ body.article .text-container > p > img {
 
 body.article figure {
     margin: rem(30px) 0;
+    page-break-inside: avoid;
```
kmcelwee, line 12: This makes sense, correct?

rlskoeser: yes

rlskoeser: although it might need to be in the print styles?

kmcelwee: It certainly works here, but if you want all print-specific styling within a `@media print` block, let me know!

rlskoeser: I moved it to the print styles file — I think it will be helpful when reviewing them or revising them to have them more centralized, in case we need to make adjustments.

gwijthoff (approved):

PR #201, opened 2021-12-07
Don't resize images larger than original

rlskoeser:
@kmcelwee would you check my logic on this? The recent geniza typeface image was not terribly large and looked pixelated when resized — I revised the figure shortcode to exclude sizes larger than the original, and always include the original (that should be redundant, but in this particular case seemed helpful, since the original was larger than the largest generated size.)


kmcelwee (approved):
Went through together

PR #216, opened 2021-12-21
213 i18n

kmcelwee:


themes/startwords/layouts/_default/list.atom.xml
```diff
@@ -44,6 +47,7 @@
             {{- end -}}
         {{- end }}
     {{- end }}
+    {{/* to here... */}}
```
kmcelwee, line 50: If you'd like to run locally immediately, remove these two sections.

kmcelwee: The error occurs because of `index site.Data.i18n.languages .Lang`. I assume that it's no longer structured as `site.Data.i18n.languages` anymore.

rlskoeser:
have adapted what I can from this pr, don't think we need to keep it open for reference anymore

PR #218, opened 2021-12-21
20 build author taxonomy

kmcelwee:
### Data template approach

I don't think it'd be too hard to add names to the yaml, we'll want to add ORCID and affiliation info anyway. 

* I assume we also want the article information as well? Do we want detail pages or we could use a `<summary>` dropdown tag? Do we want author detail pages?

- [ ] Add "sort_name" field to better sort authors

### Taxonomy approach

- [ ] Add logic to `list.html` to display ORCID and affiliation information
- [ ] Add documentation for archetype after other #217 is merged and rebasing

* Styling?
* Can information from author pages be collapsed into the list view?
* Hugo doesn't strip periods for slugs. Do we want to overwrite them?

To add metadata, we need to create folders for everyone (!) shell command to make this faster if/when we get there:
```sh
hugo new --kind author /authors/becky-rother
hugo new --kind author /authors/courtney-murray
hugo new --kind author /authors/jim-casey
hugo new --kind author /authors/justin-smith
hugo new --kind author /authors/mason-a.-jones
hugo new --kind author /authors/matthew-dudley
hugo new --kind author /authors/samantha-blickhan
hugo new --kind author /authors/shaun-a.-noordin
hugo new --kind author /authors/steffy-reader
hugo new --kind author /authors/victoria-anne-van-hyning
hugo new --kind author /authors/will-granger
hugo new --kind author /authors/gissoo-doroudian
hugo new --kind author /authors/nick-budak
hugo new --kind author /authors/rebecca-munson
hugo new --kind author /authors/rebecca-sutton-koeser
hugo new --kind author /authors/xinyi-li
```

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -11,9 +11,9 @@ enableGitInfo = true
 timeout = "300s"
 
 # disable taxonomies until we are ready to use them and have templates
-disableKinds = ["taxonomy", "term", "tag"]
-#[taxonomies]
-#  author = "authors"
+disableKinds = ["term", "tag"] # "taxonomy"
+[taxonomies]
+  author = "authors"
 #  tag = "tags"
 
 [permalinks]
--- a/config.toml
+++ b/config.toml
@@ -11,7 +11,7 @@ enableGitInfo = true
 timeout = "300s"
 
 # disable taxonomies until we are ready to use them and have templates
-disableKinds = ["term", "tag"] # "taxonomy"
+# disableKinds = ["term", "tag", "taxonomy"]
 [taxonomies]
   author = "authors"
 #  tag = "tags"
--- a/config.toml
+++ b/config.toml
@@ -11,9 +11,9 @@ enableGitInfo = true
 timeout = "300s"
 
 # disable taxonomies until we are ready to use them and have templates
-# disableKinds = ["term", "tag", "taxonomy"]
-[taxonomies]
-  author = "authors"
+disableKinds = ["term", "tag", "taxonomy"]
+# [taxonomies]
+#  author = "authors"
 #  tag = "tags"
 
 [permalinks]
--- a/content/authors/becky-rother/_index.md
+++ b/content/authors/becky-rother/_index.md
@@ -1,5 +1,0 @@
----
-title: Becky Rother
-orcid: ORCID
-affiliation: AFFILIATION
----
--- a/content/authors/becky-rother/_index.md
+++ b/content/authors/becky-rother/_index.md
@@ -1,0 +1,5 @@
+---
+title: Becky Rother
+orcid: ORCID
+affiliation: AFFILIATION
+---
--- a/themes/startwords/archetypes/author/_index.md
+++ b/themes/startwords/archetypes/author/_index.md
@@ -1,0 +1,5 @@
+---
+title: {{ replace .Name "-" " " | title }}
+orcid: ORCID
+affiliation: AFFILIATION
+---
--- a/themes/startwords/archetypes/author/_index.md
+++ b/themes/startwords/archetypes/author/_index.md
@@ -1,5 +1,0 @@
----
-title: {{ replace .Name "-" " " | title }}
-orcid: ORCID
-affiliation: AFFILIATION
----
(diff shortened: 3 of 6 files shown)


rlskoeser:
Wow, if we need author folders for everyone maybe we don't really want to use a taxonomy! 

How far would an author data file get us? i.e., put author affiliations, orcids, urls etc in a data file; add other fields as needed (maybe sort names?). I think that would let us add information to the article display if we wanted to (IDK if/how we want to, but would be good to confirm that it's feasible). I think that would also let us also generate an author list page that iterates over the names in the data file, outputs summary details, and links to articles by author.

Would also be nice if (and how hard) we can use the same data file for contributors — i.e. can we use it to link names on issue credits? What about the site masthead? (presumably that won't work in current format because it's not structured data at the moment)

kmcelwee:
@rlskoeser Working from a data file works great, we just can't have author detail pages. If we want author detail pages, then we're making directories for each author again, at which point, we might as well keep things autogenerated by the taxonomy structure, correct? I looked it up and hugo seems to be very clear:

https://discourse.gohugo.io/t/work-around-for-making-pages-from-data/13964/3
> Hugo will look for Markdown files to generate pages. There is no way around that.
There's an open hugo ticket about this: https://github.com/gohugoio/hugo/issues/5074

Using `urlize` we can easily adding a link for each author on the article template, assuming we spell names consistently everywhere. For now, I've made it an ID link to the list page (e.g. `/authors#rebecca-sutton-koeser`).

We can of course include links of articles by author, but I don't know how we might autogenerate this. Or is this something you'd expect we do by hand?

> Would also be nice if (and how hard) we can use the same data file for contributors — i.e. can we use it to link names on issue credits?

As in your name in the masthead would be linked to the list of pieces you wrote? I feel like adding a "url" property to the contributor metadata on each issue is easiest, but you'd want it centralized right? While it'd be cool to have a central roster of everyone involved and their different roles, I have a gut feeling it would introduce more complexity than it would solve.

kmcelwee:
I have absolutely no idea why tests suddenly failed after making such minor changes.

rlskoeser:
> @rlskoeser Working from a data file works great, we just can't have author detail pages. If we want author detail pages, then we're making directories for each author again, at which point, we might as well keep things autogenerated by the taxonomy structure, correct? 

I can't imagine we'll have enough information to warrant individual pages for each author — but agreed on approach, and we should make sure this is on our list of questions.

> Using `urlize` we can easily adding a link for each author on the article template, assuming we spell names consistently everywhere. For now, I've made it an ID link to the list page (e.g. `/authors#rebecca-sutton-koeser`).

I don't think names are reliable enough to use as identifiers, even if spelled consistently. Perhaps we should use slugs, or some other identifier?

> We can of course include links of articles by author, but I don't know how we might autogenerate this. Or is this something you'd expect we do by hand?

No, not by hand. I don't think that's  feasible on a generated page like this, even if we wanted to do it. I expect this to be a filtered list of pages based on type and author.

> > Would also be nice if (and how hard) we can use the same data file for contributors — i.e. can we use it to link names on issue credits?
> 
> As in your name in the masthead would be linked to the list of pieces you wrote? I feel like adding a "url" property to the contributor metadata on each issue is easiest, but you'd want it centralized right? While it'd be cool to have a central roster of everyone involved and their different roles, I have a gut feeling it would introduce more complexity than it would solve.

No, not linked to articles. I was imagining that names would be linked to profile pages or personal websites so readers could find out more information about contributors (e.g., if someone wants to know who a guest editor is and what else they've worked on).

gwijthoff:
Jumping in---without having reviewed all of the above, so this may already have been agreed on---to say that I think author information should include title and affiliation (Lead Developer, Center for Digital Humanities, Princeton University), as well as an optional URL to the author's personal website or profile page.

rlskoeser (changes requested):
This is still just a proof of concept and I'm not sure we'll ever merge this version, but the things I'd like to see:
- solution for identifiers for contributors that doesn't rely on names 
- solution for finding articles by author in the author list page

PR #219, opened 2021-12-22
204 generalize interlude

kmcelwee:


rlskoeser:
@kmcelwee weren't the interlude styles for the keyboards piece based on the interlude styles from DBV? What is proving complicated with converting it from an id to a class and pulling out the common aspects?

BTW, any idea why the background color is set in an absolutely positioned `::before` element? That seems unnecessarily complicated; wonder if it is causing some of our problems with the PDF rendering.

rlskoeser (changes requested):
I think we can do better. Or if not, would at least like to know why it doesn't work to generalize (and we should make a note).

I'd also like to see the interlude background simplified if possible (remove the absolutely positioned `::before`element).

PR #223, opened 2022-02-04
post-publication copyedits for SW2

gwijthoff:
Minor changes, no need to make new PDFs


kmcelwee (approved):
This seems fine. I assume we'll just update these on our next startwords release ?

gwijthoff:
@kmcelwee it would be nice if we can push these typo fixes to the public site now rather than waiting til SW3. @rlskoeser mentioned that we could change the dates of SW3 articles to the future, so that they wouldn't be published, and we could push other things we're staging in develop to main. At least I think that's the idea.

rlskoeser:
Given that these edits are so small, I think we should just change the PR to merge them into main and then make sure that we pull them into develop also after they're merged. 

Otherwise, we'd have to merge into develop, make a new pr with the update that changes the issue 3 pub date, then merge into main.

@gwijthoff we'll need to work with develop for the other content pages since you identified styling issues that we need to address, but it doesn't seem to be needed here.

rlskoeser:
@gwijthoff merged into `main` so the edits will be visible on the public site right away; cherry-picked the edits into `develop` so we won't lose them.

PR #232, opened 2022-03-22
content page styles

rlskoeser:
changes in this PR:
- new content page layout
- move guideline images into top-level site assets directory
- update local figure shortcode to handle image in page or site resources
- use figure shortcode in guidelines document
- apply article styles to content page layout

I noticed there's something wrong with the `about` placement in the footer on desktop but haven't investigated yet.

questions:
- what other changes need to be made to your draft documents?
- I haven't tested contextual notes yet, did you plan to add some?
- I made this a new page type, but should this be the default and the about page is the one that is special?


gwijthoff:
Answers:
- I think both style & editorial guides are good to go as they are, but we should revisit the documents this summer when we have a group meeting about SW strategic goals
- I don't have plans to add any contextual notes to the style & editorial guides, but we should probably make sure they work in case they need to be used in the future.
- Agree that this should be the default page type. However, isn't the only thing making About unique (currently `layout: page`) the indented placement of H1 and summary to make room for the nav on mobile? If so, we could resolve that and just make a single `page` layout that works for all.

Issues:
- As you said, footer placement is wonky
- Page title should be H1 but is currently (I believe) H2. Could be more padding beneath the page title and content too.
- Placement of the page title / header to the left of page content

Question: 
- Why use the [assets](https://gohugo.io/hugo-pipes/introduction/#asset-directory) directory rather than [static](https://gohugo.io/content-management/static-files/) for images?

rlskoeser:
@gwijthoff made some updates:
- tested contextual notes locally, they work properly on the guide pages
- properly scoped some of my style changes, now the footer is back to normal (but please confirm)
- renamed so that content pages are just default pages and the about page has it's own layout (it has a slightly different structure and uses the other font — I think we should keep it that way)
- added a table of contents section to the default page layout and adjusted your document structure a bit; I think this also solves our menu indentation problem: I indented the table of contents nav on mobile, the actual content falls below the nav on both of the guide pages

> Why use the [assets](https://gohugo.io/hugo-pipes/introduction/#asset-directory) directory rather than [static](https://gohugo.io/content-management/static-files/) for images?

Static content is served out as is; assets can be resized (as we do for figures, to autogenerate different sizes for different screens.)

You had a note in one of your documents about embedding a video — I think we should do it similar to what we did for the Data's Destination article, but put it in a site-wide static folder.

gwijthoff:
Right now, the page title looks like an H2. Check out the [Editorial Guidelines](https://startwords-dev-pr-232.onrender.com/guides/editorial-guidelines) page. The title is called using `{{ with .Title }}<h1>{{ . }}</h1>{{ end }}` (which is what it should be), but it looks identical to `## Meetings` in the page content. That's not currently a problem on the About page, where the "About" title is clearly bigger than all the H2's in the content.

Maybe `/startwords/themes/startwords/assets/scss/page/_global.scss` needs to be updated? Just not sure how.

gwijthoff:
@rlskoeser I'll see your 24px and raise you to 32px. There still wasn't enough distinction between H1 and H2. Now H2 is also distinct from H3. I also added some padding beneath the TOC border.

Before:
Screen Shot 2022-03-23 at 12 22 45 PM

After:
Screen Shot 2022-03-23 at 12 41 20 PM

gwijthoff:
@rlskoeser as long as you approve of the above, I think we can merge

gwijthoff:
@rlskoeser I did check simulated mobile view in Firefox and it looks good to me, but I'm not sure if there's any other "official" way to check?

rlskoeser:
> @rlskoeser I did check simulated mobile view in Firefox and it looks good to me, but I'm not sure if there's any other "official" way to check?

That's good enough!

PR #235, opened 2022-03-23
multilingual support for issue list 

rlskoeser:
changes in this pr:
- add i18n configuration to site config (adapted from Kevin's prototyping)
- copy issue index and issue 3 index english versions to spanish versions so Hugo will know they have translated content
- update issue list template to display language indicator for all languages
- update issue list template to show all issues regardless of language (i.e., spanish list displays the english versions of issues 1 and 2 instead of not listing them at all)
- styles for language indicator chips
- localize publication date display
- fixes atom feed template to not require i18n data file

notes/questions:
- The copied spanish index files are placeholders 
- I would like to implement things so we only put issue contributors on the english language index; should we make an issue for that?
- publication date doesn't get localized for english-only issues displayed on the spanish version of the issue list page; Hugo seems to switch the current language context based on the page, and I haven't figured out a way around that yet.
- I added screen-reader only text for the language indicators that says "languages available" and a label with the full name of the language; wondering if we should also have tooltips for sighted readers who aren't sure what the language chips mean?

testing checklist:
- [x] [issue list page](https://startwords-dev-pr-235.onrender.com/issues) should display a chip for languages available for that issue (EN for all three, ES for issue 3 only)
- [x] [issue list page in spanish](https://startwords-dev-pr-235.onrender.com/es/issues) should display all issues 
- [ ] date for spanish issue on spanish issue list page should be localized

design review checklist:
- [ ] review the language chip styles on the [issue list page](https://startwords-dev-pr-235.onrender.com/issues)


gwijthoff:
@gissoo I was literally just measuring the px above and below the chips to make the same comment! Good catch.

rlskoeser:
Thanks, both, for the feedback!

@gissoo I don't think you need to design tooltips — I was just planning to use browser built-in tooltip, but would appreciate help from either of you on language for screenreaders and language for tool tips. 

available in english/spanish? includes content in english/spanish? (could be different language for default site language and translated language)

rlskoeser:
one more question that occurred to me — should the order of the language chips be consistent when you switch between spanish and english? or is it ok to display the current language first?

gwijthoff:
@gissoo will probably have a better opinion on this, but it might be confusing if the order of the language chips moves. Always displaying the non-English language first, when it's available, would be a good way to avoid Anglocentrism.

> should the order of the language chips be consistent when you switch between spanish and english? or is it ok to display the current language first?

gwijthoff:
Other misc things:

> I would like to implement things so we only put issue contributors on the english language index; should we make an issue for that?

I think so, yes. Is this because each issue will probably have all articles in English, and that's the best way to pull the most comprehensive list of people for the contributors page?

> tooltips for sighted readers

Yes, that's a really good idea. I think "Available in English and Spanish" works.

> Date for spanish issue on spanish issue list page should be localized.

Febrero isn't capitalized. And on the Spanish issue list page, only the date for Issue 3 appears in Spanish. Is that what we want? Because there won't be buttons to toggle between ES and EN on the issue list (like there will be on homepage, issue page, article page), when will content appear in Spanish and when in English?

rlskoeser:
@gwijthoff I noticed Febrero isn't capitalized, but thought that was by convention — I'm using the builtin Hugo/Go date localization for that.

I couldn't figure out how to localize the dates for the issues without Spanish content — Hugo uses the context of the current page that's being rendered at that point. I wasn't sure how much it mattered; I agree it would be nice, but maybe not essential? Should we track as a separate issue and I can ask on the Hugo forums?

I think we may want to reconsider which pages we display the toggles on — they should probably be on the issue list page because it is now available in both English and Spanish, but there's no way to navigate between them. (But I suggest we tackle that in a follow up issue!)

gissoo:
@rlskoeser ok, on the order of the chips, it's always best to be consistent, I don't think we should change the order, it would confuse the users, totally agreed with Grant, this is almost always a no no in ux :D 

@gwijthoff @rlskoeser – sorry – on showing the issue/theme info in Spanish by default and reading Grant's comment, I don't think we should display it in Spanish by default – because that issue also contains an article that's only available in English – I don't think this is the same situation as displaying the article in Spanish on the homepage – also because Spanish users can see the "ES" chip on the issue list page they would still be willing to click on it to get to the ES language to read about the issue and its theme.

rlskoeser:
Updated with revisions to chip styles and to keep language order consistent (it follows site languages configured order).

@gissoo I'm not following your comment. We don't display the Spanish issue/theme info by default; we only show it when the user is viewing the Spanish version of this page. 

I thought the language chips on this page only indicators and not clickable — did I misunderstand? If someone is reading the issue list page in English and sees the "ES" for issue 3, should it be clickable to take them to that issue in Spanish?

gissoo:
@rlskoeser @gwijthoff yes, I exactly meant what Grant is saying above.

rlskoeser:
@gwijthoff I think maybe we're not all in agreement about which parts of the site are available in translation and which are not... Once you switch to the Spanish version of the site, you stay in it until you switch back. I implemented the issue list as available in translation because it was more logical to me — but I think it would be pretty easy to turn it off if we decide we don't want it. I think we should move forward with this implementation for the issue list and then refine once we have more of the site functioning in multilingual mode — either disable translation on this page, or turn on toggles everywhere.  It will be easier to see how things work — and decide how they _should_ work — as more of it is implemented.

gwijthoff:
@rlskoeser ok, that makes sense to me. It'll be much easier to visualize everything and come to a consensus once some of this functionality is working on the site. Thanks for your work on this!

PR #240, opened 2022-03-23
Localize site content in templates

rlskoeser:
in this pr:
- new english and spanish translation files in the theme `i18n` directory for site text; includes some placeholder translations from Kevin's prototyping
- use `i18n` template function throughout to display text content from the translation files based on current language
- added official Spanish translation for CC-BY license from the creative commons site
- placeholder translation for site description 
- refactor issue list credits into a partial and convert the about page to use the same structure so that we can use i18n translation for roles
- convert site config from toml to yaml


content/about.md
```diff
@@ -13,30 +31,4 @@ Each issue features two articles that speak to a common theme, feature, or conce
...
-**Board of Advisors**
-
-- Natalia Ermolaev
-- Zoe LeBlanc
-- Meredith Martin
-- Rebecca Munson (2020)
-
-{{</ wrap >}}
\ No newline at end of file
+Though we are not accepting submissions at this time, we make available for reference these working versions of the journal’s [Editorial Guidelines]({{< ref "/editorial-guidelines" >}}) and [Contributor Style Guide]({{< ref "/style-guide" >}}).
```
gwijthoff, line 34: This removes the `## Masthead` header

themes/startwords/i18n/es.yaml
```diff
@@ -0,0 +1,52 @@
+issues:
+  other: Ediciónes
```
gwijthoff, line 2: Whoa! This will make the translation so easy!

content/about.md
rlskoeser: oops! I meant to put it in the template where I'm rendering contributors, thanks for catching

Issue #247, opened 2022-03-25
feature article opening line preview needs to accommodate a wider range of text lengths


rlskoeser:

rlskoeser:
reviewed by @gwijthoff and merged into develop

PR #248, opened 2022-03-25
Preliminary work to revise feature article shape

rlskoeser:
ref #247

### DIFF
--- a/themes/startwords/assets/scss/article/_summary.scss
+++ b/themes/startwords/assets/scss/article/_summary.scss
@@ -27,16 +27,39 @@ $preview-height-xl: rem(250px);  // 1024px-
         @media (min-width: $breakpoint-m) { font-size: rem(18px); }
     }
 
+    // height must be set for shape-outside text flow;
+    // set the height based on content length
+    p { // }, p::before {
+        // position: relative;
+        // height: calc(7.6 * var(--words)) + "px";
+        // height: calc(1.4 * var(--length)) + "px";
+        hyphens: auto;
+    }
+
+    p span {
+        width: 60%;
+        // height: calc(7.6 * var(--words)) + "px";
+    }
+
+    p:before, span {
+        float: right;
+        display: block;
+        // min-height: 88px;
+        width: 100%;
+        // height: 100%;
+    }
+
+
     // carefully scale paragraph sizes with custom breakpoints to achieve
     // desired text shape. here be dragons
-    p,
+    /*p,
     p::before {
         height: $preview-height-xs;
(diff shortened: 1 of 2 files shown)


gissoo:
@rlskoeser thanks for working on this. I'll need to look deeper tomorrow but just noting that I think it would look nicer if there would be less overlap between the text and the "shadow" on mobile. (if the shadow would start a bit higher, but not as much as the one for the Spanish article, below) 

Screen Shot 2022-03-31 at 9 23 31 PM

rlskoeser:
@gissoo good point, thanks for noticing that about the shadow text! I'd like to handle revisions to the shadow text in a separate issue so we can get this one merged — are you otherwise ok with the changes in this version?

gwijthoff:
I love Hugo even more now that I know it has a function called `countrunes`.

This is brilliantly done, @rlskoeser! The ability to customize the polygon height is wonderful. And you're right, it seems that the shapes translate pretty well, without much alteration, as they scale from desktop to mobile.

For the record, I don't think hyphens are turned on. At least, I can't get any words to break as I fiddle with the shapes. But I think that's a good thing? The shapes give the reader's eye enough of a challenge as is (hopefully an interesting challenge). No need to add more visual movement with broken words too.

Go ahead and replace `len` with `runecount` and I'll review whenever you're ready.

rlskoeser:
@gwijthoff pushed the update that uses `runecount` instead of `len`, and render site for this PR has been updated. I think I already turned off the hyphenation before you looked at this.

rlskoeser:
I think I may still need to adjust the scaling of the shaping element for tablet/mobile...

rlskoeser:
FWIW, I've implemented logic to make the background text wrap exactly the same way the real text does — ready to be scaled and moved once Gissoo provides revised designs on where it should go!

gwijthoff (approved):

rlskoeser:
Ok, I adjusted the way I'm scaling the shaping element for different screen sizes. I think it's working better than it was, although it's not as good as I'd like — tablet looks the worst, but I guess that's ok since all our site analytics say hardly anyone is accessing our content on tablets. 

I think one thing complicating the layout for these elements is that Nick used a responsive grid, so the container size changes slightly every time you resize the screen — and it's really hard to make the shape of the text work with that reflowing text. And on tablet in particular, the columns are really narrow, so there isn't much room to work with to shape the text nicely.  We might think about revisiting this in future, but I think it should be out of scope to change now.

@gwijthoff or @gissoo one of you want to take a quick look at the shapes on different screen sizes and see if this looks good enough to merge?

PR #256, opened 2022-04-12
Display issue credits from default language issue

rlskoeser:
@gwijthoff  here's my solution to avoid duplicating issue contributors on translated issues:

- when rendering single issue page, if contributors are defined in issue page metadata, they will be displayed
- if no contributors are set, they will be pulled from the default language translation of that issue (i.e., for the spanish version of issue 3 they will be pulled from the english issue 3)

## review checklist:
- [x] template logic makes sense to you 
- [x] same contributors display on english and spanish versions of issue 3
- [x] readme documentation is clear


gwijthoff (approved):
Looks great!

Issue #204, opened 2021-12-09
As an editor, I want to use interlude styles on any article so that I can highlight content using startwords-specific visuals.


rlskoeser:
essays in issue 1 and 2 both use interludes as article-specific styles; we should consolidate the common styles and incorporate into main theme styles, and remove it from the article-specific styles

gwijthoff:
An important application of the interlude style I just saw in Lauren's article: she includes a screenshot of [this table](https://data-feminism.mitpress.mit.edu/pub/2wu7aft8/release/3#nobxi408tlj). It would be nice to use a markdown table instead, with the purple background of our interlude style.

Scroll down just a little to see how it currently looks in markdown on the Render site, without interlude styling: https://startwords-dev.onrender.com/issues/3/llm-limit-case/

rlskoeser:
reviewed and merged into develop in pr #271

PR #257, opened 2022-04-18
Add CSS styling for tables

gwijthoff:
The Issue 3 Klein article includes a screenshot of a table from the print edition of _Data Feminism:_

Screen Shot 2022-04-18 at 4 29 24 PM

I thought it would look better as a markdown table, rendered using our site's designs, but noticed that `<table>` styling isn't included in our CSS.

For now, I simply drew on some elements in the existing styles for article headers — sans font and 16px in two different weights - and image caption text - 16px italic serif. @gissoo what do you think?

Eventually, it would be great to further accent tables with a purple background, i.e. using the "interlude" shortcode. (Generalizing interlude styles is covered in #204 & #219.)

But for now, I thought we could style the tables like so and that would be good enough to send to Klein along with her draft, as long as @gissoo thinks everything looks alright. 

Screen Shot 2022-04-18 at 4 31 18 PM

You can view the article here: https://startwords-dev-pr-257.onrender.com/issues/3/llm-limit-case/


content/issues/3/llm-limit-case/index.md
```diff
@@ -46,15 +46,15 @@ In terms of text analysis tools more specifically, I've followed with interest [
...
+| Money and resources managed by members of minoritized groups                                                  |                 |           ✓            |
+| Data owned and governed by the community                                                                      |                 |           ✓            |
+| Quantitative data analysis “ground truthed” through a participatory, community-centered data analysis process |                 |           ✓            |
+| Data scientists are not rock stars and wizards, but rather facilitators and guides                            |                 |           ✓            |
+| Data education and knowledge transfer are part of the project design                                          |                 |           ✓            |
+| Building social infrastructure—community solidarity and shared understanding—is part of the project design    |                 |           ✓            |
+
+<table><caption>Above: Features of ‘data for good’ versus data for co-liberation, from Catherine D’Ignazio and Lauren F. Klein, Data Feminism (MIT Press, 2020), p. 140.</caption></table>
```
rlskoeser, line 57: This isn't semantic — it creates a separate, empty table with a caption.

As far as I can tell, there isn't a standard markdown format for table captions, and the format I found doesn't seem to be supported by goldmark.  It looks like doing this properly will require a shortcode — maybe we can adapt from this: https://willschenk.com/articles/2020/styling_tables_with_hugo/

themes/startwords/assets/scss/article/_single.scss
```diff
@@ -76,6 +76,39 @@ body.article article, body.page article {
...
+            font-family: $font-serif;
+            font-style: italic;
+            font-weight: 300;
+            padding-top: 1.5rem;
+            padding-bottom: 1rem;
+            width: 100%;
+            text-align: left;
+            margin: 0;
+        }
+
```
rlskoeser, line 111: this should be nested under the `table` declaration to make sure it only applies to table captions

rlskoeser (commented):
Styling looks reasonable to me, but defer to Gissoo on approving that.

Looks like this needs a shortcode to properly generate a semantic table with a caption. Not sure what that means for the text version.

I agree interlude would be nice in some cases, but think it should be opt-in because it may not always apply.

gwijthoff:
Thanks for the tips @rlskoeser!

@gissoo as long as these provisional styles are acceptable, I can send the draft to the authors

rlskoeser (commented):
The new shortcode looks good! fwiw I prefer named parameters to positional params.

You should briefly document the new shortcode in our theme readme.

gwijthoff:
Thanks! @rlskoeser: Wasn't able to switch from named to positional parameter, but we can clean that up later if you'd like.

PR #259, opened 2022-04-19
Feature/Spanish site terms

gwijthoff:
First pass at adding new translation of About page and site terms. I sent David, the translator, a couple requests for clarification, so these terms might slightly change.

One issue: changing the i18n translation key for "main" from *main* to the Spanish *principal* breaks the nav, even though it uses `{{ i18n "main" }}`. So I kept it as is "Main" seems like a pretty important word throughout the site...


gwijthoff:
Also: in the English about page metadata, Nick is listed as "Nick Budak (Technical Lead, 2020)" in the YAML. Since it's not possible to use a variable in the YAML — `Nick Budak ({{ i18n "technical lead" }}, 2020)` — I just duplicated Nick's name beneath "Technical Lead" with a year (2020), and also added RSK to "Technical", since Nick is also in both places.

rlskoeser:
> One issue: changing the i18n translation key for "main" from _main_ to the Spanish _principal_ breaks the nav, even though it uses `{{ i18n "main" }}`. So I kept it as is. "Main" seems like a pretty important word throughout the site...

Wow, thank you for catching this — I wasn't thinking when I flagged that for translation, `main` is an aria keyword that is important to keep untranslated. We should remove the `i18n` from the templates for that.

rlskoeser:
@gwijthoff just pushed a change to tweak the `nav` element so style does not key on the `aria-label`; we've used that pattern in the past because it's more semantic and doesn't require an additional attribute that's only for styling, but I think those aria-labels really should be translated. Please go ahead and add in the translation for `main` in the translation file and confirm that it doesn't break anything.

I like your solution for displaying Nick as technical lead. I wonder about displaying issue # as well / instead of year, but understand if you don't think it's worth the trouble to include that level of detail. Should we add a year (or issue #) for Kevin, also?

PR #265, opened 2022-05-23
Revise language toggles to use new background images

rlskoeser:
Have updated the language toggle buttons to use the revised background images provided by Gissoo.

View on article page: https://startwords-dev-pr-265.onrender.com/issues/3/on-spanish-parrots/ 
and on issue page: https://startwords-dev-pr-265.onrender.com/issues/3/

@gissoo the images don't seem as sharp as what is in Figma. Do you want to give me SVGs instead? Or try again with the PNGs? If you want to replace the PNGs on this branch, you can upload them here (I changed the prefix to `lang` instead of `lan`): https://github.com/Princeton-CDH/startwords/tree/chore/revise-language-toggles/static/img/icons 

But I'd be happy to upload new versions or switch to SVG if you prefer.


gissoo:
@rlskoeser I just uploaded svgs. They don't look sharp, but the spacing between them also looks off.

gwijthoff:
@rlskoeser yes, it looks to me like the text sits too low in these boxes, both in the selected and de-selected boxes, according to Gissoo's designs.

Screen Shot 2022-05-26 at 11 52 50 AM

gissoo:
@rlskoeser I think the border looks ok for now – when I inspect in the browser I see the height of the box is correct, but looks like the lower part of it is hidden/covered by something, which is why it's looking shorter than it should be, and also why the text looks too low. (on both desktop and mobile)

gwijthoff:
Thanks for the small tweaks, @rlskoeser. Can't wait to see what it looks like with three buttons for SW4: EN, JA, ZH!

PR #249, opened 2022-03-29
Preliminary author data and author list page

rlskoeser:
Please review data file and basic structure to make sure you're ok with this approach before I start converting all existing authors to this format. 

[authors list page](https://startwords-dev-pr-249.onrender.com/authors/) on render

A few notes on my approach:
- authors will be listed in articles by an alphanumeric id 
- info for authors should be added to the data file based on that alphanumeric id
- easiest for sorting if we can sort on the alphanumeric id (if we sort on other fields, we lose access to the id and would have to duplicate it; I tried some things to get around this but couldn't figure out a solution and didn't think it was worth more time)

Template logic handles:
- list authors from data file sorted by last name
- display articles by author with issue information, co-authors, tags
- list author name on detail page based on information from data file (we could display orcid id & link here if we wanted)
- link author name from article page to author list

The author list page is basically unstyled; I added a few things but not much.

PR #266, opened 2022-05-23
Authors list revisions

rlskoeser:
I started a new branch from latest develop and then pulled in the relevant changes from the preliminary author list implementation (PR #249)

I've applied Gissoo's styles and addressed the issues Grant identified (combining multiple author names with and; author names on the issue introduction page, etc.)

Review: [PR render site author list](https://startwords-dev-pr-266.onrender.com/authors/)

I need to refine how I'm handling the translated article, but I have some questions about that:
- Should we only display the article once, in English? This is what I infer from Gissoo's designs, just want to confirm before working on the implementation. I want to make sure logic for editorial introduction and translated articles match.
- Should we display translator information somewhere for articles/intros that are translated, similar to what we do for co-authors?


rlskoeser:
@gwijthoff would `_` in the author ids we're using in the data file make things more readable?

gwijthoff:
@rlskoeser answering the above:

- yes, I agree that based on Gissoo's designs, we should display the translated articles only once. The language indicator chip is enough to show that there are multiple versions.
    - related: can the tooltip for those chips on the author list page draw on i18n to say "Disponible en Español" rather than "Available in Spanish" over `ES`? Maybe it already will, and this just hasn't incorporated changes from the Spanish language terms branch.
- displaying translator info would be great. But for Issue 3, David is technically only translating the issue introduction. Gimena wrote both English and Spanish versions, and David edited them for consistency.

Also, this is a minor thing, hopefully with a minor fix: on the issue page, there's no divide between the hyperlink for each author's name. Can you make the first hyperlink extend only to the comma after the first name?

Screen Shot 2022-05-24 at 4 37 52 PM

rlskoeser:
I was trying to figure out where or if to display David as translator of the issue introduction; I feel like the site doesn't currently make it clear who translated what. Maybe `translator` isn't quite the right way to list him on the issue credits? 

Also: interesting about the "Disponible en Español" vs "Available in Spanish" ! We're using the translation string, but displaying the english version. This seems like a case where we should override the current language, yes?

FYI: it looks like we have a "read in spanish" that didn't get translated (maybe I added after you sent off)

gwijthoff:
@rlskoeser I see that missing "read in Spanish" translation. Adding it not to my list of changes in the editorial staging branch

gwijthoff:
> This seems like a case where we should override the current language, yes?

I think that's right!

gwijthoff:
@rlskoeser this is a good question about whether David is best described as "translator" in the issue credits. (Or maybe Spanish editor?) Let me give it some thought. For the purposes of this branch, I think it would be great if he could be listed as translator just beneath N & T's issue introduction.

rlskoeser:
@gwijthoff I was wondering about "Spanish editor" also!

Is it sufficient to displayed him intro translator only on the author list page or should we display it on the Spanish issue also ? (probably yes to both, but ideas where to put him on the issue page?)

gwijthoff:
@rlskoeser For the issue page, David's name could be listed at the base of the intro text, just like Natasha & Toma's. This markdown will nicely indent the second line so that it's inline with the "N" of the first line.

```markdown
— Natalia Ermolaev & Toma Tasovac
: Spanish translation by David Andrés Rivera Mosquera
```

But I'm still just using text rather than author name variables for the intro authors over in `sw3-editorial-staging`.

rlskoeser:
@gwijthoff ok, updated as follows:
- only show english version of Gimena's article on author list (with language chips)
- revise author info link display on article page (now using link icon)
- add website link & icon on author list
- add translator metadata to spanish intro and display on author list and issue page

I haven't figured out yet how to translate the "available in" language to translate into the target language instead of the current language! I think I may have to ask on the Hugo forums about that.  If the rest is good, let's merge this and not wait on that part.

gwijthoff:
@rlskoeser so this is very strange, but on the Render site there's no space between the first intro author's name and the word "and":

Screen Shot 2022-05-25 at 5 49 54 PM

Even though the space is clearly placed in the code:

`{{- if eq (add $index 1) $author_count -}} {{ i18n "and" }} {{ else }}, {{ end -}}`

And when I serve this PR locally, I get the space where you'd expect it:

Screen Shot 2022-05-25 at 5 50 26 PM

Maybe this is a Render quirk?

gwijthoff (approved):

gissoo:
@rlskoeser the designs look great. 
- On desktop, is the spacing between the article title and the lang chips greater than that of the chips and the line of additional authors below it? In the designs they are respectively 12px and 8px. They look equal in the implementation right now. 

- on mobile the lang chips labels are not vertically centered in the chip shapes, they're slightly off, towards the top.

rlskoeser:
@gissoo Thanks for reviewing. 

I checked, the spacing for the chips is 12px and 8px — but I agree with you, they look more like they are equally spaced. I think this is because there is some space between the top of the next div and where the text actually displays. (I checked, there aren't any margins or padding on the co-author text div). Do you want me to decrease the 8px slightly to account for this?

I adjusted the height of the span within the chip, I think that shifted the text placement slightly on mobile to correct the vertical centering issue you found.

rlskoeser:
@gwijthoff can we merge this in so we can get the functionality and I can open a PR for the footer update, and Gissoo and I can fine tune the layout issues in another round ?

gissoo:
Sorry for the late response! I'm just going to document responses here: 

@rlskoeser Thanks for checking – I see. Yes, let's decrease the spacing from 8px to 4px, I think that should be enough. 

For some reason the lang chip text still looks off towards the top :D – Screen Shot 2022-06-06 at 12 21 43 PM

PR #269, opened 2022-05-25
revised text shadow for feature hook 

rlskoeser:
Very excited to share the new version of the shadow text for the feature article opening lines.

Please review all issues on a variety of screen sizes:
- [homepage](https://startwords-dev-pr-269.onrender.com/)
- [issue 3 features](https://startwords-dev-pr-269.onrender.com/issues/3/#features)
- [issue 2 features](https://startwords-dev-pr-269.onrender.com/issues/2/#features)
- [issue 1 features](https://startwords-dev-pr-269.onrender.com/issues/1/#features)

I've also added some space after the third feature on issue 3 since there are no snippets.

### DIFF
--- a/themes/startwords/assets/scss/article/_summary.scss
+++ b/themes/startwords/assets/scss/article/_summary.scss
@@ -95,14 +95,14 @@ $preview-height-xl: rem(250px);  // 1024px-
         left: 0;  // attach to left of viewport; keep fixed with text vertically
         width: 100%;
         color: $light-purple;
-        opacity: 0.1;
+        opacity: 10%;
         pointer-events: none;
 
         // use scale for sizing to ensure text shape matches visible summary
         // scale relative to initial font size of 1rem (18px)
 
         // on mobile, position 21px above corresponding summary text
-        // desired font size is 19px; original mobile font size is 16, so 1.18
+        // desired font size is 19px; original mobile font size is 16, so 1.18x
         transform: translate(0, -21px) scale(1.18);
         // set transform origin relative to left so that scaling will not move off-screen
         transform-origin: left top;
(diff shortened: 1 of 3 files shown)


gwijthoff:
This is gorgeous! Great work @rlskoeser and @gissoo — the new text shadows add so much more character to the site!

One issue I noticed is that these text shadows allow horizontal scrolling beyond what you'd expect on mobile, whereas before, the text was firmly placed in the center and the user could only scroll vertically. E.g.:

https://user-images.githubusercontent.com/5580840/170374988-95833791-8e7e-490a-981f-201202f7e421.MP4

gissoo:
@rlskoeser Thank you for working on this!!! Looks great!

- I was able to replicate the issue Grant is showing above on firefox and chrome browsers for mobile view. 
- I also noticed on chrome and firefox there is a lot of extra space in the middle of the preview text: (issue 2 looks fine on chrome):
Screen Shot 2022-05-26 at 10 02 36 AM
Screen Shot 2022-05-26 at 10 02 50 AM

on safari the extra space is between the preview text and the article title that's on the right side of the page for all issues except issue 2! (so like the below): 
Screen Shot 2022-05-26 at 10 05 30 AM


If you can't replicate this, ignore me, it's probably because I need to update my browsers again. I don't see this problem on Grant's screen recording.

rlskoeser:
The horizontal scrolling was caused by the secondary shadow text running off the edges of the page, and it only happens on iOS.

I've added some style adjustments that I think should correct this problem — but I'm only testing in the browser, could one of you confirm that it's fixed?

@gissoo I think what you're seeing is that the shaping element doesn't work perfectly on all screen sizes, because the container that includes the feature articles is responsive. I have the shaping element configured to resize somewhat based on screen size, and we can now override the height for articles where the default calculation doesn't work — so if this is happening on any screen sizes you think are important, we can tinker with it some. But I'd prefer to get this merged and revisit in another round if necessary.

PR #271, opened 2022-05-25
Generalize interlude styles

rlskoeser:
- generalized `interlude` class to make it part of startwords theme style for use in any essay
- customizations specific to Data Beyond Vision or the keyboards piece remain in the article-specific styles as overrides/extensions
- revised table shortcode to use named parameters and support a `class` parameter (although ended up not using it)
- added two interludes to Lauren's essay to test the styles (the table and her three assertions)

review essays with interludes:
- [x]  [Klein](https://startwords-dev-pr-271.onrender.com/issues/3/llm-limit-case/), issue 3
- [x] [keyboards](https://startwords-dev-pr-271.onrender.com/issues/2/strangers-in-the-landscape/), issue 2
- [ ] [data beyond vision](https://startwords-dev-pr-271.onrender.com/issues/1/data-beyond-vision/), issue 1


rlskoeser:
@gwijthoff when I refactored the interlude styles, the embedded keyboards by default were displaying at a wider width than the rest of the text content — which I actually like better. It seems like it's more consistent with how we've displayed content in interludes before, and it gives the keyboards more room. Take a look and see what you think. If you prefer it to look like it does on the published site, I'm willing to adjusting the width so they match the other content.

FWIW, I noticed when I was checking the revised styles that a couple of the later embedded keyboards are actually too big on mobile. It's the same way on the published site and I didn't see an easy fix. I don't know how much this matters, maybe it's fine.

gwijthoff:
@rlskoeser Nice! I'm okay with the keyboard text entry boxes being wider. Another thing I noticed though is that the line height in these interludes is now larger than the typical body text line height. That ends up adding a lot of extra, unnecessary vertical space to the keyboard interludes, and makes the "see / touch" interlude in DBV look less like verse, which feels more right for that example.

rlskoeser:
Wow @gwijthoff, good catch, thank you! the paragraph styles were in place but the selector wasn't specific enough, so the interlude paragraphs were getting default article paragraph styles instead.

I've corrected that and also removed the bold formatting on the list in the first interlude in Lauren's essay.

See what you think.

Issue #274, opened 2022-05-26
As a reader, I want to see DOIs for all versions of translated articles so that I know what languages the article is available in and can find and read them.


rlskoeser:

rlskoeser:
reviewed merged into develop in PR #275

PR #275, opened 2022-05-26
Display all DOIs for translated articles

rlskoeser:
ref #274


gwijthoff:
I like how that looks. Minor request: could you also add both language DOIs to the .txt version of the article? 🙏🏻

gwijthoff (approved):

PR #276, opened 2022-05-26
Localize text version of articles

rlskoeser:
When I added the translatotin DOIs to the text version of the article I noticed we had a couple of things that weren't translated or localized and should be. This adds them.


gwijthoff:
Hmm, "Spanish" isn't appearing next to the related DOI!

Screen Shot 2022-05-26 at 2 39 44 PM

Two other things I'm noticing: shouldn't the comma come after the link icon, rather than separating the name and the link, since the link belongs to the name? Similarly, there's no comma or "and" separating the last two authors.

Screen Shot 2022-05-26 at 2 42 45 PM

And, can we put a space between the name and the link icon, just so it doesn't become confused with the letter glyphs in the author's name?

Screen Shot 2022-05-26 at 2 42 08 PM

rlskoeser:
weird about the language names! it worked for me locally, wonder what is different... could it a be a translation file discrepancy, I wonder?

I fought a bit with the placement and spacing of the link icons, and giving them enough space in the different scenarios (directly after an author name or after a comma). I wasn't sure whether the link should come before or after the comma — your logic on that makes sense, and moving them might simplify the layout problems I was dealing with.

There's a comma between the last two authors — we've always just had commas and no "and" for author titles on the article page. It looks like there is one in your screenshot — am I missing something?

gwijthoff:
@rlskoeser omg there _is_ a comma there! Sorry, doing too many things at once. And you're right, I don't think there's an "and" required between the last two names.

But yea, I think the link should come next to the author's name and before the comma.

gwijthoff:
@rlskoeser I think displaying in the target language makes sense. I'm thinking about the way Wikipedia lists the other languages a given article is available in on the bottom left of each article. Each wiki is listed in its target language.

rlskoeser:
@gwijthoff ok, revised doi language labels and author list comma/info link spacing and order

gwijthoff:
Perfect, thank you @rlskoeser!

Issue #268, opened 2022-05-23
As a reader, I want an easy way to discover the author list page, so that I can see it exists and access it from any page on the site.


rlskoeser:
## dev notes
revise footer based on Gissoo's designs to add author list link

rlskoeser:
merged & available on the dev site

PR #277, opened 2022-05-26
Update footer config & styles to add Authors page

rlskoeser:
ref #268

in this PR:
- revise menu configuration in site config to add Authors after About
- adjust footer styles per Gissoo's designs to accommodate displaying both Authors and About

Note: I thought it was better for the Spanish site menu to be consistent with the English site, so I included the English site Authors link in the Spanish menu. Open to other ideas if you prefer something else (but if we only have one item in the menu the styles will probably need some adjustment).

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -37,6 +37,10 @@ menu:
     - name: About
       url: about/
       weight: 20
+    - name: Authors
+      url: authors/
+      weight: 30
+
 
 Author:
   name: Center for Digital Humanities at Princeton
@@ -79,6 +83,10 @@ languages:
         - name: Acerca de
           url: /es/acerca-de/
           weight: 20
+        - name: Authors
+          url: authors/
+          weight: 30
+
 
 markup:
   goldmark:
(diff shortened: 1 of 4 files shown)


gwijthoff:
Cool! I agree with you about the Spanish / English handling and think it works best as is. However, can you make the font weight 400 rather than 300 for those "About" and "Authors" page links? 400 would match the "issues" nav text at top left, and the dictionary definition text.

rlskoeser:
I defer to @gissoo on the font weight, I think I've implemented it as she designed.

@gissoo I'm getting a lighthouse error about the menu item tap targets not being large enough, although for some reason it's only happening on the 404 page.  (Maybe ignorable if we think it isn't a real problem.)

gissoo:
@rlskoeser @gwijthoff I'm fine with matching the weight of the links in the footer with the "Issues" in top nav – the size will need to stay smaller though – my only concern is whether the weight will create problems with spacing in desktop/mobile due to the small space in the footer. 

I think it's fine to ignore lighthouse error for now – if it turns out to be problematic beyond the 404 error page I can revise the height of the footer container.

PR #278, opened 2022-05-26
Update notes to replace location instead of change

### LINKED ISSUE #273: contextual notes pollute browser history
If you're reading through an article and read the notes as you go, then you want to use the browser back button to go back to the issue list to read a different article, you have to go back through every note you read because they are in the browser history as distinct urls.

The contextual note js should be updated so that it changes the location (so you can still link to a specific note) but doesn't pollute the browser history — maybe when the note is closed.

rlskoeser:
fixes #273

### DIFF
--- a/themes/startwords/assets/js/notes.js
+++ b/themes/startwords/assets/js/notes.js
@@ -31,6 +31,15 @@ class ContextualNotes {
             ContextualNotes.showNote();
         }, false);
 
+        // bind to footnote reference click to manage browser hsitory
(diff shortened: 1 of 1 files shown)


rlskoeser:
Happily, there was a simple fix for this bug. @gwijthoff please review and confirm. fwiw, I've tested on Firefox desktop and mobile (android) and Vivaldi (≈ Chrome) desktop.

gwijthoff:
Sweet! I tested in Safari and Firefox and navigating back ignores the footnote links that have been clicked.

PR #281, opened 2022-06-15
Add language toggles to home and about pages

rlskoeser:
review:
- [x] language toggle on homepage https://startwords-dev-pr-281.onrender.com/
- [ ] language toggle on about page https://startwords-dev-pr-281.onrender.com/about/


gwijthoff:
@rlskoeser the language toggles look great! One problem I noticed though is that the nav bar (homepage / issues) is no longer sticky on the about page: https://startwords-dev-pr-281.onrender.com/about/

gwijthoff:
@rlskoeser one more thing: the English About page no longer has the "Masthead" h2, though the "Encabezado" h2 is present on the Spanish version.

rlskoeser:
wow, @gwijthoff good catch!

rlskoeser:
@gwijthoff  I added the missing English translation for "masthead" and changed the styles and placement for the language toggles on the about page. This is ready for a re-check.

gwijthoff:
@rlskoeser a question on this one we don't necessarily have to worry about now: in the future, will language toggles appear on the homepage for English-only issues?

gwijthoff:
@rlskoeser one lingering problem: for the Spanish about page when viewed on mobile, the nav doesn't hide when you scroll down. It does hide on the English about page.

rlskoeser:
> @rlskoeser a question on this one we don't necessarily have to worry about now: in the future, will language toggles appear on the homepage for English-only issues?

Yes. In Hugo terms, the homepage is the site root and so it should normally be available in all supported languages. The other option I considered is that the homepage link could always take you to the English language homepage, but it seemed weird to me that going to the homepage would kick you out of your chosen language. We can revisit when we have more issues with translation if necessary.

gwijthoff:
@rlskoeser sounds good. Thanks for clarifying. Excited to see how this language logic works for future issues!

rlskoeser:
> @rlskoeser one lingering problem: for the Spanish about page when viewed on mobile, the nav doesn't hide when you scroll down. It does hide on the English about page.

Wow, another good catch! The `nav.js` was still keying on the `aria-label` attribute, which we dealt with for CSS and determined should be translated. Just pushed a revision that should address.

PR #300, opened 2022-08-01
Update keyboards piece to provide access to code within the article

rlskoeser:
We've talked about startwords pieces including code snippets, but the one piece we've published so far that was technical enough to do so, we don't show the code! I think we should make the code more easily available in the article, since people will be much more likely to look at it there (vs. at least two clicks otherwise).

in this pr:
- new short code (& accompanying styles) for displaying code from a file in a collapsed details/summary element
- document the short code in theme readme
- use the new short code to embed a viewable version of the source code in the keyboards article for each embedded keyboard

----- 

to review:
- [x] review the file changes in this pr
- [x] look at the article with code embedded in the piece on render site: https://startwords-dev-pr-300.onrender.com/issues/2/strangers-in-the-landscape/

### DIFF
--- a/themes/startwords/README.md
+++ b/themes/startwords/README.md
@@ -262,6 +262,18 @@ Either `pdf-img` or `youtube_id` should be specified.
 
 [view source](layouts/shortcodes/video.html)
 
+### source code
+
+Use to display highlighted source code from a file in a details/summary element.
+Path should be relative to content directory.
+
+Example use:
+```
+{{< source_code source_file="path/to/code.html" format="html" >}}
+```
+
+[view source](layouts/shortcodes/source_code.html)
+
 ## Author metadata
 
 New authors should be added to the author data file with name, title, affiliation, and optional website URL and ORCID. Identifiers in the author data file should be in `PascalCase` based on last name first, so that names can be sorted based on those identifiers. Authors of articles and issue introductions should be listed in page metadata using the same identifier as the data file, for linking the content and displaying the author's name properly.  If an author's sort name cannot be programmatically determined from their display name, it should be set explicitly in the author data file.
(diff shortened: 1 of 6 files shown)


themes/startwords/README.md
```diff
@@ -262,6 +262,18 @@ Either `pdf-img` or `youtube_id` should be specified.
...
+
+Use to display highlighted source code from a file in a details/summary element.
+Path should be relative to content directory.
+
+Example use:
+```
+{{< source_code source_file="path/to/code.html" format="html" >}}
+```
+
+[view source](layouts/shortcodes/source_code.html)
```
gwijthoff, line 275: @rlskoeser Is it just me, or could the "view source" link right beneath the `source_code` example use be confusing? At first glance, I still thought the "view source" hyperlink was _part_ of the "Example use." But again, that might just be me, looking at one shortcode documentation in isolation. Someone reading the entire theme README would probably recognize that all of the shortcodes contain a "view source" link at the bottom of each section.

gwijthoff: Only thing I can think of is to change the "view source" hyperlink at the end of each shortcode section in the theme README to something like "view this shortcode", or "view shortcode source". Shortcode sourcecode? 🤭 Maybe this is a nonissue.

rlskoeser: @gwijthoff OMG it _is_ kind of confusing! I didn't even think about that, just followed the pattern from how we documented the shortcodes.  (oh no, could be "view source code shortcode source code " 🤣 ) 

I think isn't quite as bad on the [rendered version of the readme](https://github.com/Princeton-CDH/startwords/tree/chore/include-keyboards-code/themes/startwords) as it is in the markdown. I'm open to revising the label, but if we do we may want to revise the others as well.

gwijthoff: @rlskoeser agreed. Let's leave it as is for now. We can always revisit later.

Issue #288, opened 2022-07-26
missing space between introduction author names


gwijthoff:
For some reason, I can't replicate this bug when serving locally. But on the Render site for the develop branch, at the base of the introduction, there's no space between the first author name and the word "and" [in English](https://startwords-dev.onrender.com/issues/3/) or "y" [in Spanish](https://startwords-dev.onrender.com/es/issues/3/).

rlskoeser:
confirmed resolved on dev render site by changes in #305

PR #305, opened 2022-08-03
X2 space fix

gwijthoff:
The current double space solution is still bothering me. (Ref #288.) There's no way this code...

`{{- if eq (add $index 1) $author_count -}}&nbsp;{{ printf " %s " (i18n "and") }} {{- else -}}, {{- end -}}`

...should produce this result!

Screen Shot 2022-08-03 at 3 42 10 PM

And, for some reason, that zombie space only rises from the dead on Render, or the production site, but not when serving locally. I know we still can't figure out how to remove the zombie space, but in this branch I replaced `&nbsp;` with an invisible, zero-width space character, copied and pasted from this [list of whitespace characters](https://qwerty.dev/whitespace/).

Let's see if it works on the Render site that this PR creates.


gwijthoff:
@rlskoeser Better, but still not fixed with the zero-width space.

Screen Shot 2022-08-03 at 4 01 42 PM

rlskoeser:
@gwijthoff the PR isn't showing any files changed, is this a glitch on GitHub's part?

gwijthoff:
omg Rebecca I created the branch and the PR but didn't push my changes

gwijthoff:
@rlskoeser ok ready for review

rlskoeser (approved):
Looks much better on render site.

Thanks for documenting it! FWIW, I checked and the invisible space is visible (sort of) in my code editor, so that's helpful.

PR #313, opened 2023-05-18
Handle missing tweets

rlskoeser:
Another of the tweets referenced in the community essay in issue 2 has been deleted, which is breaking our build. 

Came up with a simple solution to just display the text from the tweet and not that it's been deleted; see what you think.

### DIFF
--- a/content/issues/2/creating-and-recreating-virtual-community/index.md
+++ b/content/issues/2/creating-and-recreating-virtual-community/index.md
@@ -71,7 +71,9 @@ Launching our social media campaign early on helped attract attention to the eve
 {{< tweet user="BrigField" id="1360976957337456641" >}}
 {{< tweet user="DGershenowitz" id="1361014486589706249" >}}
 {{< tweet user="hralperta" id="1361039589914382344" >}}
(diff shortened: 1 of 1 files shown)


gwijthoff:
@rlskoeser so it looks like this is a fix just for those deleted tweets / accounts, rather than a programmatic fix in case additional people leave the platform, right?

Seems like this is a good solution for now.

rlskoeser:
> @rlskoeser so it looks like this is a fix just for those deleted tweets / accounts, rather than a programmatic fix in case additional people leave the platform, right?
> 
> Seems like this is a good solution for now.

Yes, that's right. Fortunately we had the content available in our plain text version, although we also have it in older versions and the PDF. It would be nice to have something a bit more durable, wouldn't it? I've been wondering if Hugo actually needs to rebuild pages when the content hasn't changed...

There's also probably a way to capture and display the twitter data a bit differently, but IDK how much of a priority that is for us - doesn't seem relevant for any of our upcoming articles or snippets, but let me know if you think it will be relevant anytime soon.

Thanks for reviewing & merging!

Issue #316, opened 2023-06-01
revisit article text template approach


rlskoeser:
would it work to use `.txt` templates for all the relevant parts of an article (including shortcodes) instead of using the regex approach we're using now?

Issue #303, opened 2022-08-03
some characters still appearing as ASCII codes in TXT


gwijthoff:
We thought this was a Render bug, but smart quotes and smart apostrophes (I think that's all?) still render in ASCII — `&#34;` and `&#39;` — in the TXT files on the production site.

See https://startwords.cdh.princeton.edu/issues/3/mapping-latent-spaces/index.txt

rlskoeser:
@gwijthoff I tried a few different patterns to our plain text replacement config file but none of them made any difference. I have thoughts about a different way to implement the plain text, maybe we can revisit how we're doing that at some point. (And/or reassess whether it's worth keeping — I'm not sure how to assess whether people are using this version or find it useful, but it does add a fair amount of overhead to the technical and editorial work.)

gwijthoff:
@rlskoeser gotcha. Let's revisit some other time.

PR #329, opened 2023-09-05
Try out alternate approach to article text output

rlskoeser:
ref #316

Preliminary work to explore an alternate approach for generating the plain-text version of articles.

I've added `.txt` templates for our figure and pull quote shortcodes, and updated the Underwood article to use them. As you can see, we no longer have to duplicate the plain text version of the figure in the article source, we can generate it from the template.

If we go with this approach we'd have to rewrite or add regexes to run against html instead of markdown (unless we can find a better approach than that).  From my experimentation, I think that this approach might also solve the character bug in #303 

Please review the new text templates, the corresponding changes to the Underwood article, and the text version of that article on the Render PR site and provide feedback on whether you think this direction is worth pursuing.


gwijthoff (commented):
Brilliant solution to replacing those janky regexes! I love having the pull quotes in the .txt output. And that would be great if this approach solves the #303 character bug. When you say "rewrite or add regexes to run against html instead of markdown," do you mean find a way to translate the `<i>html<i>` to `*markdown*` output in the .txt file? We could strip out all html markup, but I feel like it's important to maintain the markdown in the .txt since it's both human readable and machine executable, and keeps intact all of the authorial intent in the piece.

rlskoeser:
@gwijthoff yes, was thinking we'd need to write new regexes to convert the html to whatever we want the text to look like — similar to what we do now, converting markdown footnotes into [NOTE 2]. I wish there was a better option than regexes, but at least maybe text versions of shortcodes will improve things some.

gwijthoff:
@rlskoeser I agree that this would be a marginal improvement over our current TXT output and is worth pursuing. Not a must-have for Issue 4 in case your capacity is tight next week, but it would definitely be great.

rlskoeser:
closing this PR for now since we don't have capacity to investigate further; will keep the branch and the issue in the icebox to track the initial work

PR #335, opened 2023-09-08
Styles for SW4 UL figures 2 and 6

rlskoeser:
updated styles for SW4 "An Unnatural Language" figures 2 and 6

- implemented figure 6 based on Gissoo's designs
- tweaked styles for figure 2 based on Grant's comments

I cut the explanatory text before figure 2 because it seemed like it was written for us working on the figures, not for the readers. If we want some of that content included, can it be consolidated into the figure caption?

The table for figure 2 doesn't work well on mobile, it's too large; it might be better if it stacked on mobile, but it does make semantic sense as a table and I'm not sure I want to rework all the html and styles to figure out how to make it stack; doing that would also make it less accessible to screen readers.


gwijthoff:
@rlskoeser Yea, I just noticed the mobile issue for the table in Fig. 2 as well. Can we make our table shortcode side-scrollable when the content is too wide? (For some reason I thought they were already.) We'll need to do something about that for [Xin's piece](https://startwords-dev-pr-335.onrender.com/issues/4/toward-deep-map/) anyway too -- check out his fig. 3 on mobile.

Looking at [Gissoo's design](https://www.figma.com/file/MEt3gfKX4N3WTLPbLGBbGC/Startwords?type=design&mode=design) for fig. 6, I can see that stacking the two text blocks vertically will prove easier wrt mobile designs. But is there a way to shrink the padding between the text block and the figcaption?

Screenshot 2023-09-11 at 12 21 43 PM

rlskoeser:
> Looking at [Gissoo's design](https://www.figma.com/file/MEt3gfKX4N3WTLPbLGBbGC/Startwords?type=design&mode=design) for fig. 6, I can see that stacking the two text blocks vertically will prove easier wrt mobile designs. But is there a way to shrink the padding between the text block and the figcaption?

I was trying to follow Gissoo's designs. But maybe we only need that extra space on mobile and not on tablet?

rlskoeser:
> Can we make our table shortcode side-scrollable when the content is too wide? (For some reason I thought they were already.) We'll need to do something about that for [Xin's piece](https://startwords-dev-pr-335.onrender.com/issues/4/toward-deep-map/) anyway too -- check out his fig. 3 on mobile.

@gwijthoff I see a way to do this. I think it needs to be opt-in; easiest would be to specify a class when you use the table shortcode, and we can document the option in the readme. What class-name would make sense to you? horizontal-scroll? h-scrollable? side-scroll ?

Do you want that change on this branch or a separate one?

gwijthoff:
> I was trying to follow Gissoo's designs. But maybe we only need that extra space on mobile and not on tablet?

@rlskoeser If it's an easy fix to not have that padding between the table text and the figcaption on desktop and mobile, then let's remove it. Otherwise we can close and merge this PR. I do think it's distracting how big this particular interlude box is on desktop, but it's not a major deal.

rlskoeser:
@gwijthoff I'd like to merge this and revisit spacing for figure 6 when I incorporate the audio. Ok with you?

Let's make an issue for that, and then you can add any notes for how you'd like me to implement it. Gissoo's design didn't include a figure caption, and I'm not sure I got the spacing exactly right.

rlskoeser:
I'm going to merge this. Refinements will be tracked on #347, please add any comments or additional guidance there.

PR #336, opened 2023-09-08
Update babel/core versions to address deprecation warning

rlskoeser:


rlskoeser:
lighthouse check is failing because of the console error on the font reference that got messed up; I'll wait until I hear back from the docraptor folks about variable font support so I know which way to fix this

PR #339, opened 2023-09-11
Update print styles for docraptor and add python script for generating PDFs

rlskoeser:
in this PR:
- revised print styles to work with docraptor; PDFs have been generated from this branch and reviewed by Grant
- new python script to generate PDFs using the docraptor API and docraptor python package
- modes for single article PDF _and_ for generating PDFs for all feature articles in an issue; IDK how far this gets us towards #6 but it's a nice step that I think will save me time!
- usage instructions at the top of the python script

question:
- is documentation in the script sufficient, or should this be documented in the readme?


rlskoeser:
@quadrismegistus could you look over the new python script and see if it all makes sense / if you have any concerns? If you're up for testing it out and making sure my documentation is enough, that would be helpful too.

scripts/create_pdf.py
```diff
@@ -0,0 +1,123 @@
...
+a test PDF is the default behavior for this script. When you are ready to
+create a final PDF, use the `--no-test` flag to turn test mode off.
+
+To generate PDFs for all the feature articles in one Startwords issue,
+use the `--issue` flag and run with the issue url, e.g.,
+`create_pdf.py https://startwords.cdh.princeton.edu/issues/3/mapping-latent-spaces/ --issue`
+
+PDFs will be named according to current Startwords PDF naming conventions.
+
+"""
```
quadrismegistus, line 24: Nice, very clear documentation, and it's cool it accepts (already legible) URLs as input

scripts/create_pdf.py
```diff
@@ -0,0 +1,123 @@
...
+                # test documents are free but watermarked
+                "test": test,
+                "document_type": "pdf",
+                "document_url": url,
+                # "javascript": True,
+                "prince_options": {
+                    "media": "print",  # @media 'screen' or 'print' CSS
+                },
+            }
+        )
```
quadrismegistus, line 55: So this just returns the raw binary for the output PDF? Nice

quadrismegistus: It also reads in the URL directly. Is it investigating the DOM in some way? Looking for paragraph tags etc?

scripts/create_pdf.py
```diff
@@ -0,0 +1,123 @@
...
+    if not response.status_code == requests.codes.ok:
+        response.raise_for_status()
+
+    # parse the xml and find the links to feature articles
+    issue = BeautifulSoup(response.content, "html.parser")
+    # find the links for the articles
+    # limit to links under /issues/ (no author links, DOIs, etc)
+    article_links = issue.find(id="features").css.select('a[href^="/issues/"]')
+    # snippets are under features; get a snippet link list to exclude them
+    snippets = issue.css.select('.snippets a[href^="/issues/"]')
```
quadrismegistus, line 80: Nice -- we put at id="features" in the HTML there for this or similar purposes?

scripts/create_pdf.py
```diff
@@ -0,0 +1,123 @@
...
+            continue
+        article_parsed_url = parsed_issue_url._replace(path=link["href"])
+        article_url = article_parsed_url.geturl()
+        # filenames are similar to urls;
+        # convert /issues/N/title-slug/ to startwords-N-title-slug.pdf
+        output_filename = "startwords-%s.pdf" % "-".join(
+            link["href"].strip("/").split("/")[1:]
+        )
+        print(f"\nCreating PDF from { article_url }\n\t➡️ { output_filename }")
+        create_pdf(article_url, output_filename, test=test)
```
quadrismegistus, line 96: Nice, so this just calls the other function for each article

scripts/create_pdf.py
```diff
@@ -0,0 +1,123 @@
...
+    parser.add_argument("-o", "--output", help="Output filename for PDF")
+    parser.add_argument(
+        "--no-test",
+        action="store_false",  # turn test mode off when this parameter is specified
+        default="true",  # testing is true by default
+        help="Turn off the test flag and generate real PDFs",
+    )
+    parser.add_argument("--issue", action="store_true")
+
+    args = parser.parse_args()
```
quadrismegistus, line 113: I don't know if you've used `click` before, it simplifies argparsing, but it's another dependency and this works fine

themes/startwords/assets/scss/print.scss
```diff
@@ -13,32 +13,31 @@
   }
 
   @top-right {
-    content: element(page-header);
+    content: flow(page-header);
```
quadrismegistus, line 16: what does `flow` do?

quadrismegistus (approved):
Looks great! very elegant

scripts/create_pdf.py
rlskoeser: It's using the generated html + the print styles and then somehow applying them to created a paginated pdf that uses the paged media css spec.

There's an option to turn on javascript if you need it for rendering;  I don't think we need it for any of our SW PDFs so keeping that disabled for now.

themes/startwords/assets/scss/print.scss
rlskoeser: It's part of the paged media CSS spec - it takes an element out of the regular document and inserts it into a space for a running header or footer.  

Here's the docraptor documentation for it, which is a little more readable & targeted than the full paged media spec https://docraptor.com/documentation/article/1067094-headers-footers

I'm still not sure why I had to switch from `element` to `flow` but this is what they have in their documentation and it worked the way I wanted. (We use this to put the SW logo in the top right corner, DOI in the bottom left, etc.)

rlskoeser:
Since Grant has reviewed and signed off on the PDFs generated by this new code and now that Ryan has reviewed the python code, I'm going to get this merged into develop so we can start working with the new print styles and PDF generation as we work on finishing the remaining figures.

PR #344, opened 2023-09-12
Make body text font-weight 400, not 300

gwijthoff:
Currently, the body text of _Startwords_ is set to a font weight of 300. This looks a little spindly and light, and is difficult to read for extended passages of text. Most body text on the web is set to 400, which is the OpenType specification for "normal / regular" text. (300 is light, 500 medium, 700 bold, etc.)

As discussed with @rlskoeser, who recalled that @gissoo's original designs set body text at 400, this change would bump up all body text to 400. Other bits of text are retained at 300 as originally intended for emphasis and variation (like in the metadata block at the top of articles).

@rlskoeser if you still agree with this change, please verify that this PR addresses every place that body text was set to 300, and that you approve how the new body text looks in articles.

### DIFF
--- a/themes/startwords/assets/scss/_fonts.scss
+++ b/themes/startwords/assets/scss/_fonts.scss
@@ -121,7 +121,4 @@
     // harmonize sizing with font for Latin characters
     // on desktop, main font should be 18px and Noto Serif TC should be 16px
     size-adjust: 89%;
-    // we're using regular font weight (400), but want it displayed
-    // with the main body text, which as currently font-weight 300
-    font-weight: 300;
 }
--- a/themes/startwords/assets/scss/_global.scss
+++ b/themes/startwords/assets/scss/_global.scss
@@ -47,5 +47,5 @@ dd {
 /* Use Noto Serif TC for anything explicitly labeled as Chinese language */
 *[lang=zh] {
     font-family: "Noto Serif TC";
-    font-weight: 300;
+    font-weight: 400;
 }
--- a/themes/startwords/assets/scss/article/_interlude.scss
+++ b/themes/startwords/assets/scss/article/_interlude.scss
@@ -8,7 +8,7 @@
     padding: 35px 0;
 
     color: #E0CEFF;
-    font-weight: 300;
+    font-weight: 400;
     font-size: 16px;
     font-family: 'IBM Plex Serif', 'Palatino', serif;
 
--- a/themes/startwords/assets/scss/article/_single.scss
+++ b/themes/startwords/assets/scss/article/_single.scss
@@ -22,6 +22,7 @@ body.article {
             }
 
             .number {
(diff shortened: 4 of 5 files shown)


themes/startwords/assets/scss/_fonts.scss
```diff
@@ -121,7 +121,4 @@
     // harmonize sizing with font for Latin characters
     // on desktop, main font should be 18px and Noto Serif TC should be 16px
     size-adjust: 89%;
```
rlskoeser, line 123: I think we need to keep the font-weight, just adjust it to match the font style where this font will be used.

```suggestion
    size-adjust: 89%;
    font-weight: 400;
```

rlskoeser:
> who recalled that @gissoo's original designs set body text at 400,

If this is what I said, I misspoke - what I meant to say was that I noticed the designs for the new Noto CJK font used 400 weight and I wondered why it was different.

I just checked and the original designs for desktop use 300 weight, so the implementation does follow the original design.

I don't know if I'm just used to it, but the weight doesn't bother me and I haven't found it difficult to read. I'd like to ask  Gissoo to see if she remembers why she went with that weight (I wonder if it was for contrast with headings, which are 500?), and what the ramifications would be for changing this.

rlskoeser:
@gissoo in your original designs for SW you set article font weight to 300. @gwijthoff has flagged this as unusual, and I noticed that in the latest design for the Noto CJK you used 400 weight in your Figma mockups.

Can you remember why you chose this weight, and let us know what concerns you would have about shifting body text from 300 to 400? Will this be problematic with other font weights used throughout the site?

gissoo:
@rlskoeser @gwijthoff  thank you for flagging this!! It was not intentional, it was as mistake. I just revised [the mockups on this page ](https://www.figma.com/file/MEt3gfKX4N3WTLPbLGBbGC/Startwords?type=design&node-id=4634%3A10388&mode=design&t=GLaDfirKWsdpwRbA-1) – so the article font weight (ibm plex serif) is light/300, and changed all the text pieces in noto serif cjk to be 300/light.

rlskoeser (commented):
I don't think we should merge this. If it's important to you to assess and adjust, I think we would need someone to look at the typography and weights across the entire site, and we don't have the capacity to do that right now.

gissoo:
@rlskoeser @gwijthoff  Sorry I didn't get the point of this issue at first as I got caught up reading Rebecca's comment and my mistake on the designs for issue 4. I can assess the change of type from 300 to 400 on the site if you'd like. I think we chose 300/light because it looked comfortable to read and not as dark and thick as the regular for this particular font on the article page.

rlskoeser:
Thanks, @gissoo. I wondered if it was a font-specific choice. Reassessing the font weights is not in scope right now.

Issue #341, opened 2023-09-12
Enable editor to specify position of word break for issue list page


gwijthoff:
Currently, the one-word titles of our issues on the [issue list page](https://startwords.cdh.princeton.edu/issues/) break after a certain number of characters (I believe). However, with Issue 4's name, PALIMPSESTS, there is a trailing 'S' on a line of its own. It would be useful for the editor to be able to specify where the word should break, so that each issue title breaks neatly. In this case, for example, we could specify `4` to get

```
PALI
MPSE
STS
```

or `6` to get

```
PALIMP
SESTS
```

gwijthoff:
Addressed in #348

PR #348, opened 2023-09-14
Add & document theme word wrap override parameter #341

rlskoeser:
- adds a new optional parameter to override theme wrapping width
- add new parameter and document in issue index archetype template
- document the new parameter in theme readme
- use the new parameter for issue 4


gwijthoff:
This is great! Thank you @rlskoeser!

Issue #346, opened 2023-09-14
Implement the visualization for figure 7 that communicates the Jingdian Shiwen annotation 


gissoo:
@rlskoeser Here is the finalized version on [desktop](https://www.figma.com/file/MEt3gfKX4N3WTLPbLGBbGC/Startwords?type=design&node-id=4855%3A2736&mode=design&t=rtmzpZvDIRBbs1Kh-1) and on [mobile](https://www.figma.com/file/MEt3gfKX4N3WTLPbLGBbGC/Startwords?type=design&node-id=4855%3A2895&mode=design&t=rtmzpZvDIRBbs1Kh-1). 

Question for you: Since I have customized the height/length of the underlines based on dimensions of content would you want me to upload all of the underlines as unique assets? Or just upload the longest versions? 
like the below ...[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdXFhIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--1e8892873fb0dde91b53ddf588db266f287bdf5a/Screenshot%202023-09-14%20at%201.51.19%20AM.png)

Note: 
- on desktop and mobile there are auto layout structures through which you can inspect the spacing and margins. The way it's structured might be a bit strange on desktop as there are nestings but it's the way that made the most sense to me. Please let me know if you have any questions. There is a unified 24px horizontal spacing on desktop between each word/character. 
- on mobile because of the way I have structured the auto layout I had to cheat and put the off white backgrounds (#fefefe) behind the text, so here is the specs for top and bottom margins of the content within the off white boxes: 8px.  

## development work
- [x] restructure html
- [x] styles for new layout
- [x] annotation border image styles
- [x] Noto Sans CJK

rlskoeser:
@gissoo thank you for creating the issue and documenting all of this.  Please export the smallest version of the underline images that can be used to repeat/tile and get the desired pattern; i.e., one dot with appropriate spacing, one squiggle, etc.

rlskoeser:
@gissoo CSS border supports dots! `border-style: dotted;` see https://developer.mozilla.org/en-US/docs/Web/CSS/border-style

It's possible to generate a wavy border using a radial gradient; I don't know how hard to translate your designs to that, but here's one tool that will help you generate the CSS: https://css-generators.com/wavy-shapes/ - but it does look quite like your wave pattern. 

[...]

PR #351, opened 2023-09-14
Preliminary markup and styles for figure 7 #346

rlskoeser:


rlskoeser:
@thatbudakguy this is a preliminary implementation of Gissoo's designs for your figure 7. Could you review to check that I've restructured the html properly and have all the correct terms matched up?  (Comments on the approach are welcome if you have time & interest.)

rlskoeser:
@thatbudakguy @gissoo ok, I've made some progress on the styles but not fully working. When I adjusted the styles for mobile I messed up something with the padding for some of the underlines on desktop and it's late on a Friday so I'm going to leave it for now.   

@thatbudakguy any ideas how to do the full-width white and gray stripes on desktop? Do you think a background gradient would work?

thatbudakguy:
re: stripes, haven't tried this out, but could you use the same approach as the purple interludes? i.e. a `::before` pseudo-element that is set to `width: 100vw;` (looks like)

thatbudakguy:
see #352 for my suggestions.

gwijthoff:
This looks phenomenal! Thanks for your work on figure 7, everyone.

gwijthoff:
@thatbudakguy sorry, getting a little lost in all the versions here. Can you resolve the conflict above then merge the PR? I'm not sure which should be used: the HTML & caption text from `develop` or from `sw4-feature1-fig7`.

thatbudakguy:
I don't have write access to this repo, so unfortunately I can't do that...I'm happy to give it a shot if you grant me access as a collaborator, or @rlskoeser could resolve them.

PR #352, opened 2023-09-16
Implement figure 7 using underline styling

thatbudakguy:
Here's my crack at figure 7. I wanted to try using underline styling exclusively to draw the borders; the trick for mobile is positioning a pseudo-element that hosts the underline and making it `writing-mode: vertical-lr` so that the underline becomes a side border.

Note that I didn't match @gissoo's styles completely: empty circles are not a supported underline style. Instead, I kept the colors and tried to use the 5 underline styles the platform makes available, since it'll render close to the same across all browsers and doesn't need an image to do so (solid, dashed, dotted, wavy, double).

## desktop

Screenshot 2023-09-16 at 14 16 44

## mobile

Screenshot 2023-09-16 at 14 16 44

### DIFF
--- a/content/issues/4/unnatural-language/index.md
+++ b/content/issues/4/unnatural-language/index.md
@@ -413,14 +413,14 @@ Only with such a digitized version of the *Jingdian Shiwen* and related texts ca
 <div class="graphic segment">
   <div class="source" lang="zh">本又作磨</div>
   <div class="notes">
-    <div class="translation">Edition[s] also write it [摩] as “磨”</div>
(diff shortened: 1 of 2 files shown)


rlskoeser:
@thatbudakguy thank you so much for taking this on. Either I didn't know or I'd forgotten that underlines included a wavy style option. I think this is fantastic and much prefer using browser underline styles to tiling images. Your screenshots look great.

I'll merge into my branch so we can review with render and see if anything else is needed to merge into develop.

gissoo:
@thatbudakguy thank you!! These look good, they look distinct enough from one another.

PR #353, opened 2023-09-16
Update figure 6 based on Gian's notes

thatbudakguy:
* Don't use IPA slashes for Middle Chinese
* Use Baxter's transcription system
* Fix the highlighting on "fish"
* Indicate language on spans to fix font usage

We might want to revisit the faded highlights on the transcriptions...it's kind of hard to tell which thing is being highlighted, especially because the font that gets used to render IPA characters looks bolder than the regular one:

Screenshot 2023-09-16 at 15 20 00

### DIFF
--- a/content/issues/4/unnatural-language/index.md
+++ b/content/issues/4/unnatural-language/index.md
@@ -376,18 +376,20 @@ While earlier dictionaries primarily glossed glyphs by providing similar-soundin
   <aside class="card">
   <header>Say <em>east</em> in Middle Chinese</header>
   
-  <p class="large"><span class="tek">東</span> /tuŋ/</p>
-  <p class="large"><span class="tek">德 +</span> 紅 = 東</p>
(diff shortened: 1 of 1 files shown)


rlskoeser:
> the font that gets used to render IPA characters looks bolder than the regular one

@thatbudakguy from your screenshot, this looks like a fallback font sizing issue. I'm not sure how we determine which font the IPA characters are coming from, or if it's worth adding a font declaration so we can apply a relative sizing here. 

Would switching to a glow to highlight instead of a fade to de-emphasize help? Or could we apply an article-specific %-based sizing to the faded portions?

rlskoeser:
I'm going to go ahead and merge in the corrections so we have the updates.

thatbudakguy:
I think you're right that it's a fallback issue — using the glow might help here too, I hadn't thought of that!

PR #354, opened 2023-09-16
Add a shadow to the deep zoomable viewer

thatbudakguy:
I was noticing that it was hard to determine where the borders of the zoomable viewer were when the image had a lot of white in the background. This just adds a simple box-shadow to make it "pop" a bit more. Just a suggestion 🤷🏻 

Screenshot 2023-09-16 at 15 25 50

Screenshot 2023-09-16 at 15 24 33

### DIFF
--- a/themes/startwords/assets/scss/article/_single.scss
+++ b/themes/startwords/assets/scss/article/_single.scss
@@ -234,6 +234,8 @@ body.article article, body.page article {
 
     .deepzoom {
         width: 100%;
+        box-shadow: rgba(0, 0, 0, 0.35) 0px 5px 15px;
+
         /* limit height to avoid scroll traps on mobile */
         max-height: 50vh;
         @media (min-width: $breakpoint-xs) {


rlskoeser:
@gwijthoff this looks good for Nick and Gian's piece; any concerns about how it will look for the other deepzoom images (and possible PDFs?). Should we make it an article-specific style, or an opt-in style, or just apply to everything?

gwijthoff:
At a quick glance, I don't see any problem with the deepzoom shadow in other articles. It might make sense to make it opt-in, in case any unforeseen issues arise with future images. But we can just make it global for now, and revise the code to make it opt-in later. Or would you prefer to do it now, @rlskoeser?

rlskoeser:
Pulled locally to check. It looks ok on Data Beyond Vision (although not necessary / doesn't add much) and it kind of messes up the Weaving as Interface piece (although that one is admittedly a bit different).  It's much more important to have it on deep-zoom images with light backgrounds. I'll tweak it to make it opt-in but available site wide.

PR #360, opened 2023-09-19
Add & configure Noto Sans TC

rlskoeser:
this is the last task for #346 but it applies to other aspects of this issue

Noto Sans TC has been configured the same way that Noto Serif TC was, and should be applied everywhere sans-serif fonts are in use


thatbudakguy (approved):
seems good — note to myself that I think we want to make the large-format characters in excerpts from ancient texts (notably figures 3, 4, and 7) serif because it feels kind of anachronistic to display them in a "modern-looking" sans font. but elsewhere in figures I think the sans is more readable, especially for smaller-sized chinese text.

rlskoeser:
> seems good — note to myself that I think we want to make the large-format characters in excerpts from ancient texts (notably figures 3, 4, and 7) serif because it feels kind of anachronistic to display them in a "modern-looking" sans font. but elsewhere in figures I think the sans is more readable, especially for smaller-sized chinese text.

Yeah, interesting - that makes sense to me. Happy to approve and merge a PR whenever you have updates to tweak the fonts for your figures.

PR #361, opened 2023-09-20
Add audio to figure 6

rlskoeser:
https://startwords-dev-pr-361.onrender.com/issues/4/sonorous-medieval/#fig6


content/issues/4/sonorous-medieval/style.css
```diff
@@ -225,6 +226,50 @@ cite.book::before {
...
+  border: 1px transparent;
+}
+button.play:focus-visible {
+  border: 1px dashed white;    
+  outline: none;
+}
+
+button.play:after {
+  mask-image: url("/img/icons/phosphor/play-circle-fill.svg");
+  mask-size: cover;
```
thatbudakguy, line 248: chrome 116 didn't recognize these, but did offer to autocomplete the `-webkit-mask-image` property, so we might just need the vendor-prefixed version here too.

Screenshot 2023-09-20 at 09 31 11

thatbudakguy (changes requested):
the buttons unfortunately don't display in chrome, but I think there's an easy solution (noted on the CSS). i had trouble understanding [the full picture on caniuse.com](https://caniuse.com/?search=mask-image), but it seems like using the vendor prefixed version might fix it.

Screenshot 2023-09-20 at 09 31 01

one other thought — if it's not too much effort, what if we listened to the [`playing` and `ended` events](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio#events) emitted natively by `<audio>` to swap the play button for a pause button? i doubt the user would ever actually have time to pause, but i found it a little weird that nothing at all happened after i clicked the button (and there was a noticeable delay while the audio loaded and then played). maybe some other form of UI change to indicate "playing" would suffice too.

rlskoeser:
> the buttons unfortunately don't display in chrome, but I think there's an easy solution (noted on the CSS). i had trouble understanding [the full picture on caniuse.com](https://caniuse.com/?search=mask-image), but it seems like using the vendor prefixed version might fix it.

Whoops, thanks for flagging! Somehow I thought we were using autoprefixing, but that must be a different app.

> one other thought — if it's not too much effort, what if we listened to the [`playing` and `ended` events](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio#events) emitted natively by `<audio>` to swap the play button for a pause button? i doubt the user would ever actually have time to pause, but i found it a little weird that nothing at all happened after i clicked the button (and there was a noticeable delay while the audio loaded and then played). maybe some other form of UI change to indicate "playing" would suffice too.

I wondered about that - the audio clips are so short I didn't think a pause was really necessary, but I didn't think about the possible delay when the server is remote.  Gissoo didn't provide any state designs so I just extrapolated a bit with existing colors and designs elsewhere on the site.  Do you think it's enough if there's a pause button to indicate that the clip is playing, or if we add a pause button will it need to be clickable?

thatbudakguy:
> Do you think it's enough if there's a pause button to indicate that the clip is playing, or if we add a pause button will it need to be clickable?

I honestly don't know...maybe Gissoo has an opinion on it? My instinct is that it's so short that being able to actually pause it doesn't make sense, so it's fine to not have it be interactive while playing.

thatbudakguy (approved):
looks & works great in chrome now!

rlskoeser:
@thatbudakguy Grant was concerned about how much space there was in an earlier version of this figure - I didn't change the layout or spacing, but think it's better now with the buttons for the audio clips. Are you happy with how the figure looks in terms of layout / spacing / design?

thatbudakguy:
just checked the spacing/etc. on desktop and mobile and it looks fine to me.

PR #363, opened 2023-09-21
SW4 figure tweaks

rlskoeser:
- apply side-scroll class to table in figure 2; tweak spacing and make row headers sticky
- adjust spacing / alignment for figure 6 with text
- use hugo to minify article-specific js


thatbudakguy (approved):
spacing & alignment changes look good

PR #365, opened 2023-09-21
Complete figure 4

thatbudakguy:
- Add note to caption about glowing text
- Add note to caption about asterisks/reconstructions
- Dim non-rhyming lines in table
- Adjust font sizing for desktop vs mobile
- Make chinese characters use serif font
- Add some aria markup for indicating rhyming sounds

### DIFF
--- a/content/issues/4/sonorous-medieval/style.css
+++ b/content/issues/4/sonorous-medieval/style.css
@@ -69,7 +69,7 @@ cite.book::before {
   background-color: var(--purple); /* match interlude background */
 }
 
-/* figure 3 (provisional) */
+/* figure 3 */
 
 #fig3 {
   font-family: var(--font-sans);
@@ -160,15 +160,27 @@ cite.book::before {
   text-decoration-style: wavy;
 }
 
-/* figure 4 (provisional) */
+/* figure 4 */
+
+#fig4 .caption {
+  line-height: 1.25;
+}
 
 #fig4 table {
-  min-width: 50%;
+  width: 100%;
+  max-width: 32rem;
+  margin: 0 auto;
 }
 
 #fig4 table caption {
-  font-size: 24px;
+  font-size: 1rem;
   caption-side: top;
+  line-height: 1.25;
+
+  @media (min-width: 768px) {
+    font-size: 1.25rem;
+    max-width: 32rem;
+  }
 }
 
 #fig4 .stanza-header {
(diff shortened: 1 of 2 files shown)


content/issues/4/sonorous-medieval/style.css
```diff
@@ -84,15 +83,26 @@ cite.book::before {
...
 }
 
 #fig4 table caption {
-  font-size: 24px;
+  font-size: 1rem;
   caption-side: top;
+  line-height: 1.25;
+
+  @media (min-width: 768px) {
+    font-size: 1.25rem;
```
rlskoeser, line 104: I think setting a max-width on the caption fixes the table alignment on desktop.

```suggestion
    font-size: 1.25rem;
    max-width: 32rem;
```

rlskoeser (changes requested):
Lighthouse is complaining about `aria-description`; did you mean to use `aria-label`? Seems like your description is short enough for a label.  

The adjustment of the table width for desktop didn't look like it was working; I'm suggesting a change to set a max-width on the caption as well as the table, which I think resolves the problem.

The other changes look fine to me.

thatbudakguy:
I set `aria-description` because I didn't want to override the label (which is just the content of the sound), but just add some extra context for non-sighted users (i.e. that it's rhyming). VO reads it as "rhyming sound, [transcription]" instead of just "[transcription]" which seemed better to me 🤷🏻 

That said, getting anything useful at all out of these transcriptions might be a losing battle...SRs really mangle them (some of them, inexplicably, seem to be read in a russian-sounding voice??) and there's no `lang=` attribute code that you can use to hint at what voice algorithm because...IPA and other systems aren't a language! so I'd be OK doing something else (or nothing) too.

rlskoeser:
> That said, getting anything useful at all out of these transcriptions might be a losing battle...SRs really mangle them (some of them, inexplicably, seem to be read in a russian-sounding voice??) and there's no `lang=` attribute code that you can use to hint at what voice algorithm because...IPA and other systems aren't a language! so I'd be OK doing something else (or nothing) too.

Yeah, I was wondering how this would be made accessible... I wonder if an extended description would be a different way to approach this?

thatbudakguy:
That's probably the right call! I'll draft something and get it in this PR.

thatbudakguy:
ok, I kept it simple: just added some additional SR-only explanation to the `<caption>` for both table describing the key insight in more detail.

rlskoeser (approved):

gwijthoff:
@thatbudakguy the new caption is _perfect_, thank you!

Issue #368, opened 2023-09-28
images smaller than recommended minimum size get stretched


rlskoeser:
we're not generating derivatives bigger than the original, but the css styles result in the smaller images being displayed at full content width and low resolution

gwijthoff:
Resolved in #370

Issue #342, opened 2023-09-12
Enable user to specify image size smaller than full-width


gwijthoff:
Currently, our image shortcode by default sizes an image to the maximum width of the screen or the text container, I believe. But several images in the Issue 4 feature "Toward a Deep Map" are low quality, especially Fig. 1. It would be useful to have a way to specify the display of a smaller size image for lower resolution image files.

gwijthoff:
Resolved in #370

PR #370, opened 2023-09-28
Add optional max image width to figure shortcode

rlskoeser:
proposed solution to #368 

My first thought was to automatically set a width based on the image resolution for images under a certain size threshold; when I did that, I discovered that the second image the [Underwood piece](https://startwords.cdh.princeton.edu/issues/3/mapping-latent-spaces/) is low enough resolution that it would be affected by this, and I thought it was probably better to let content editors control this behavior.

I also documented the max height figure parameter, which was not listed in the readme but present in the shortcode


rlskoeser:
I guess this also at least partially addresses #342 - I was originally thinking of implementing more semantic styles like half article width and adapting the pull quote left/right styles to allow using that with smaller images, but maybe we don't need all of that

gwijthoff (approved):

PR #372, opened 2023-09-28
adjust print styles for SW4 figures 

rlskoeser:


rlskoeser:
@thatbudakguy I've made several adjustments to the figures in your article so they'll display better in the PDF. Wherever I've adjusted sizes, I've kept them proportional - although we may want to adjust a bit more to sizing and spacing based on how the PDF is turning out. (I'll share a test PDF so you can look.)

Something is going on with figure 7 that DocRaptor can't handle. Their documentation says they support flexbox but not grid; I thought maybe the nested flexbox was a problem, but I've tried several alternatives that look fine in Firefox print preview but still don't display well in DocRaptor. I just submitted a help request; hopefully we'll hear something soon...

@thatbudakguy please review the style changes and the PDF I'll send; if you want to make any adjustments yourself you can use the new script for generating PDFs with DocRaptor.

cc @gwijthoff

rlskoeser:
@gwijthoff for print, we may want to put table captions at the top instead of the bottom. DocRaptor wraps tables pretty nicely and repeats the table headings, but if the table wraps the caption comes at the bottom of the first page, not at the end.

Our default styles are to avoid page breaks within interludes; I've turned this off for a couple of Nick and Gian's figures but we may need a bit more fine-tuning for the flow of content in the PDF of their article.

rlskoeser:
update:
- I figured out part of the problem, this is CSS and not SCSS so all media queries need to be top-level; I think the browsers tolerate that but DocRaptor does not; figure 7 is looking much better than before now 
- I don't think docraptor supports text-underline-offset; not sure there's anything we can do about that
— docraptor either doesn't support `width: min-content` or interprets differently than the browsers; the translation text in figure 7 isn't wrapping to the width of the chinese it accompanies; I have a support request asking for advice on this but we may not hear anything back useful or hear anything in time.  If we have to, I suppose we could force it to the width we want for docraptor only....

gwijthoff (approved):

rlskoeser:
completed in #372

PR #385, opened 2024-10-08
Add plausible analytics template and configuration

rlskoeser:
This PR adds a template and config for Plausible analytics tracking, so we can make a more informed decision about switching from Google Analytics to Plausible. This structured the same way we added it to CDH django applications, with intentionally minimal configuration.

FWIW I saw some warnings / errors on the google analytics template when I was testing this, hopefully something out of date on my side that you've already dealt with for the new issue.


rlskoeser:
replaced by #386

